PR #2, opened 2016-04-07
increased abstraction: set cmake variable OPCUA_TOOLKIT_LIBS based on…

ben-farnham:
Hi Piotr: what do you think? 

Better to make a TOOLKIT_LIBS variable and use that in the target_link_libraries step (and reuse in other target_link_libraries steps where other exe's are created - i.e. googletest programs)

… release/debug build configuration. Allows to use same cmake variable in other link targets, googletest executables for example


pnikiel:
Yep - I think it is a very good idea.
Go ahead with the merge.

PR #3, opened 2016-04-22
Python improvements on the error handling

damiron:
The purpose of this changes is to add additional checks that ensure that all the dependancies needed by quasar.py are there before starting the execution, and if something is missing, show a clear message explaining what is missing.
This changes were requested by the jira issue OPCUA-583.

Added a new python file, externalToolCheck.py, in which the different external tools are checked in order to verify that all the external dependencies are met.
This file was added in original_files.txt, and the functionality is called from quasar.py.
The check will always verify the following external dependancies when quasar.py is executed:
- saxon9he.jar in the Design folder
- java
- Astyle
- kdiff3
- CMake
- make (Linux) or msbuild (Windows)
- xmllint

Apart from this, if quasar.py generate diagram is called, the check will also verify GraphViz. On the other hand if quasar.py doxygen is called, it will also verify Doxygen.

### DIFF
--- a/FrameworkInternals/addressSpaceGenerators.py
+++ b/FrameworkInternals/addressSpaceGenerators.py
@@ -20,7 +20,6 @@ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND 
 '''
 
 import os
-import subprocess
 import platform
 from transformDesign import transformDesignVerbose
 asPath = "AddressSpace" + os.path.sep
--- a/FrameworkInternals/automated_build.py
+++ b/FrameworkInternals/automated_build.py
@@ -20,10 +20,11 @@ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND 
 '''
 
 import os
-import subprocess
 import platform
 import __main__
 from generateCmake import generateCmake
+from externalToolCheck import subprocessWithImprovedErrors
+from externalToolCheck import getVcvarsallPath
 
 def findFileRecursively( topdir, target ):
 	for dirpath, dirnames, files in os.walk(topdir):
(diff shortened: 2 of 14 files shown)


pnikiel:
Hi Damian,

I was wondering if it wouldn't be much simpler to catch the exception coming from subprocess.call ?

damiron:
Hello Piotr

> Hi Damian,
> I was wondering if it wouldn't be much simpler to catch the exception coming from subprocess.call ?

Yes, this will be simpler, but it will be very hard to separate errors caused by not having the tool from real problems. Doing it this way, we always do the simplest operation possible with the tools, so we can ensure that if this fails, it means that the tool is not accessible.

pnikiel:
Hello Damian,

In my opinion, here is what stands out from your implementation:
* the implementation ignores exceptions. Exceptions is THE way to know that something failed.
* instead the implementation has a mechanism which checks for call success in completely remote place, again - ignoring info from exceptions and rather checking for return code.

I made the test myself to be sure that I'm advising a solution that actually works.

Could you please have a look at following python code:
```

import subprocess

try:
    #subprocess.call(["/this_executable_for_sure_doesnt_exist"])
    subprocess.call(["/tmp"])
except Exception as e:
    import pdb
    pdb.set_trace()
```

run this and then when pdb prompts, investigate the exception type:

(Pdb) p e
OSError(13, 'Permission denied')     ( this is what happens when you execute /tmp - first line commented)

(Pdb) p e
OSError(2, 'No such file or directory')      ( this is what happens when you uncomment the first line from this example)

The documentation for standard exceptions can be found here:
https://docs.python.org/2/library/exceptions.html

In addition, the proposed implementation duplicates the paths to varied tools used by quasar providing no coherency (i.e. when we change the name of code formatter then the invocation has to be updated both in the factual scripts and in the new file added to the implementation). This unfortunately makes it more difficult to manage and more prone to errors.

damiron:
Hello Piotr. Sorry for taking a long time to answer, notifications come to my personal email and I somehow skipped this one.
You bring a good point to the table. Your solution is definitely more elegant. When designing my solution I was thinking about also checking the tools that will be later used by CMake. However upon further inspection I realized that all the operations done by CMake come trough the python scripts anyway, so they would be fulfilled trough the return code of subprocess too.
I'll take a look at your approach.

damiron:
Ok, so comming back to this, as I finally got time to do so.

I am going to implement your approach. However since the "Test all the dependencies in a row" functionality is already implemented and working, what if we leave it there, as another quasar.py command? I can definitely see it being useful when setting quasar in a new environment.

pnikiel:
Hi Damian,

not sure how to read your question because you closed the request. Is it still open?

damiron:
I am reworking this branch into a better and improved version. Don't worry, wait until I create another pull request ;)

pnikiel:
OK. So let me comment on another aspect. I wouldn't mind that the feature to check tools is there provided that it is implemented in consistent and easy to manage way.

To be specific wrt this pull request:
* the list of tools got duplicated. This is of course counter productive when it comes to managing the functionality. What If there is just a dictionary defined that stores the mapping:

(tool_type, platform) -> path

Where tool type is one of XSLT, Astyle, Kdiff etc, and platform is one of Linux, Windows?

Having done so. the tool check is automated by iterating over all mapping items for given platform, and also in the scripts to generate device logic, address space etc we could remove if/else windows/linux to obtain the path...

damiron:
Hello Piotr, I don't really what paths you mean. The only case of an absolute path on the commands and on the test is the script vcvarsall.bat, which is as you say hard coded, and not nice.
However all the rest of applications that we use are just accessed trough a simple command, no path needed.

Implementing what you suggest in a clean, generic, and multi-platform way will not be an easy task. If we take a look at the dependency tests themselves, the commands I use for validating the existence of a program are wildly different and most of them only valid for their specific case. As regards for the device logic itself, often we are using the same command but with different parameters. Even things like the python module sub process don't operate the same way in Linux and Windows, as you would expect it to (something that costed me a few hours of frustration today.) 

Taking all of this into account, implementing what you suggest won't be easy, and definitely out of scope for this branch, which was supposed to be for a small improvement, and it is dragging on quite a bit already.

I already had my eye on the issue with the vcvarsall.bat script, which in the current state it is hardcoded in 2 different places. While the decision to call this script using an absolute path comes from mimicking the normal usage of the script, I have centralized it to be hard coded only once at least.
Please take a look at the pull request that I will do in a while, but I don't really have time to develop the system that you suggest, specially because I think it will be way more complex than what you think.

pnikiel:
My biggest concern is that when you read the file FrameworkInternals/externalToolCheck.py it is terribly redundant and it looks as a big copy-and-paste work. Redundancy never pays off, especially for managing the code.

 Also I see no consistency between hard-coded data in the file (i.e. that Java executable is called java) and what we then have in other scripts.

Imagine at some moment we switch from default java to java-1.8. 

At how many places we have to substitute java to java-1.8? It looks like that at least 2, and maybe more. This is what I refer to by saying redundant.

PR #6, opened 2016-05-02
[OPCUA-611] Fix for debug builds sometimes stopping with abort() call

ben-farnham:
Fix for OPCUA-611, initialize uninitialized variable - stops call to abort() on startup. Only affects debug builds.


pnikiel:
Thanks!

PR #7, opened 2016-05-11
Bugfix in shutdown.cpp (file provided by unified automation)

damiron:
Fixed a bug in shutdown.cpp (file provided by unified automation) in which, in some version of windows, if a set of specific conditions is met, windows.h will not be included and the file will fail to compile, since it is needed to resolve the variable MAX_PATH.
In the bugfix, windows.h will be included if the macro WIN32 exists (ergo, the system is a windows machine).
Bug defined in jira issue OPCUA-616.


pnikiel:
thanks, merging!

PR #8, opened 2016-05-23
Python improvements on the error handling (and more)

damiron:
Set of python improvements, mainly (but not only) oriented at having better error handling, and dependency checks. Changes include:

- Command dependency_check added to quasar.py and to quasarGUI.py. This command will call the function checkExternalDependencies inside the (newly created) file externalToolCheck.py. This will go trough all the dependencies checking if they are available, and printing messages accordingly. The programs that are checked are the following: Saxon, Java, Kdiff, CMake, Compiler (either make or msbuild), xmllint, astyle (optional dependency), GraphViz (optional dependency), DoxyGen (optional dependency).

- Improved error control when using the subprocess python module. Now instead of calling subprocess directly, the functions subprocessWithImprovedError or subprocessWithImprovedErrorsPipeOutputToFile are called. This functions are defined in the file externalToolCheck, and call subprocess as expected, but managing the exceptions to detect when the command that you are trying to run is non existent or non accessible. In case that you need to pipe the output of a command to a file, subprocessWithImprovedErrorsPipeOutputToFile should be used instead, as the way this was implemented in the past (using the switch shell=True) is unsafe and does not control exceptions. The effect of using this trought all the python scripts is that when a external dependency is used, while it is not available, a proper error message is shown now to the user.

- Astyle changed from compulsory dependency to an optional one. In the past, if astyle failed to execute (because of an error on astyle itself, or because it was not present) the execution of quasar.py just stopped at that point. Now this is not the case anymore, and if there is a problem executing astyle for whatever reason, a warning is printed, and execution of quasar.py goes on. This should have been like this from the beggining, as Astyle only provides asthetical changes into the autogenerated code.

- Introduction of the file commandMap.py, where a map of all the different commands used for calling the external dependencies is stores. So if for example java is called trough the command java9 in the next version, you only have to update this information in commandMap.py, as all the other files will access this information trough the map.

### DIFF
--- a/FrameworkInternals/addressSpaceGenerators.py
+++ b/FrameworkInternals/addressSpaceGenerators.py
@@ -20,7 +20,6 @@ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND 
 '''
 
 import os
-import subprocess
 import platform
 from transformDesign import transformDesignVerbose
 asPath = "AddressSpace" + os.path.sep
--- a/FrameworkInternals/automated_build.py
+++ b/FrameworkInternals/automated_build.py
@@ -20,10 +20,11 @@ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND 
 '''
 
 import os
-import subprocess
 import platform
 import __main__
 from generateCmake import generateCmake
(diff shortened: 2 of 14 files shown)


ben-farnham:
files missing, will be re-submitted when files added to branch

pnikiel:
Hi Damian,

could you upmerge your branch ? Would be easier to watch the diff with 
regard to most current state.

As for the topic itself, could we meet for 1 hour sometime this week 
(Thursday, Friday) or next week?

cheers p

PR #12, opened 2016-06-17
Python improvements by Piotr and Damian

damiron:
Set of python improvements, mainly (but not only) oriented at having better error handling, and dependency checks. Changes include:

• Command exteral_tools_check added to quasar.py and to quasarGUI.py. This command will call the function checkExternalDependencies inside the (newly created) file externalToolCheck.py. This will go trough all the dependencies checking if they are available, and printing messages accordingly. The programs that are checked are the following: Java, Kdiff, CMake, Compiler (either make or msbuild), xmllint, astyle (optional dependency), GraphViz (optional dependency), DoxyGen (optional dependency).

• Improved error control when using the subprocess python module. Now instead of calling subprocess directly, the functions subprocessWithImprovedError or subprocessWithImprovedErrorsPipeOutputToFile are called. This functions are defined in the file externalToolCheck, and call subprocess as expected, but managing the exceptions to detect when the command that you are trying to run is non existent or non accessible. In case that you need to pipe the output of a command to a file, subprocessWithImprovedErrorsPipeOutputToFile should be used instead, as the way this was implemented in the past (using the switch shell=True) is unsafe and does not control exceptions. The effect of using this trought all the python scripts is that when a external dependency is used, while it is not available, a proper error message is shown now to the user.
• As a side effect of this, return codes were completely dropped in the internal python functions, and all the different problems and issues are now handled trough exceptions.

• Introduction of the file commandMap.py, where a map of all the different commands used for calling the external dependencies is stores. So if for example java is called trough the command java9 in the next version, you only have to update this information in commandMap.py, as all the other files will access this information trough the map.

• Introduction of the file quasarCommands.py where the relationship between commands (such as generate device for example) and the functions they call is defined. This information was previously stored in quasar.py

• When using quasarGUI.py a message was always printed to let you know the quasar.py equivalent command. This was done individually in each function, but now it is done in a generic way straight away from quasarGUI.py, using the information stored in quasarCommands.py, making the code needed for this much cleaner and smaller.

• Several small refactoring improvements, like doping unnecessary imports, merging similar functions, correcting typos in the error messages, and improving them overall, improving the information provided for certain commands and other small changes.


pnikiel:
Hi, 
I already quite carefully reviewed it so I'm all for for this merge.
Is anybody else willing to review? Otherwise we could just go ahead.

PR #17, opened 2016-08-11
Open62541 backend support bf

ben-farnham:
Hi All,

It's a big one but I believe it's ready. This branch comprises the following:
1. Dual UA backend support (Unified Automation and open62541), thanks to Piotr and Michael for their hard work: I've just curated their changes into a single merge-able-to-master branch (I hope).
2. Meta works as expected for Unified Automation and open62541, including discovering LogIt component Ids (from LogItComponentIds.h) at server start up.
3. Quasar address space items are added to NS2.
4. Remaining certificate validity period is displayed in Meta (or, if the certificate is not in the expected place or with the expected name a suitable message is displayed). This last point was not strictly related to open6 support, but since open6 support and Certificate stuff both involved Meta changes it seemed easier to do it on the branch to avoid merge conflicts etc.

A final note: Please can we review and merge this ASAP - it's a big branch and there seems to be a strong correlation between a branch age + size and merge difficulty. At time of writing GitHub is telling me this branch can be merged automatically - excellent.

Cheers,
Ben


pnikiel:
Hi Ben, All,
my comments:

1) thanks for making Meta work for open6, not sure yet how this is done? Looks you deferred invocation of linkChildNodeToParent, is that correct understanding? In any case it is done very cleanly, thanks.
I've briefly tested it (in open6, not yet in UA) and it looks OK.

2) Some data in Meta when using open6 backend looks questionable, i.e. connectedClientCount always says 0. I understand why, however could I ask you to consider spawning an improvement ticket that, when open6 is used, those fields stay in "Uncertain" or even "BadUnsupported" OPC-UA status??

3) I've looked at the new feature: remaining certificate validity time.
I was thinking about how an end-user would use this feature in production environment?
In order to profit from the information, I think that remaining val time shall be some numeric field (expressed in i.e. days, hours or so) such that subscribed consumer of data (i.e. WCCOA project) would have an alert on this field.
Of course I have nothing against text based representation (like now) but it is not very useful outside "manual" debugging.
Also, what I noticed is that the status stays good even when there is no valid certificate (the string content is OK though). Perhaps non-OK status would be more appropriate.

4) General comment: certificate validity check probably deserves 2 or 3 lines of documentation.

5) What happens when this function fails? 
SSL_use_certificate_file( m_ssl, m_certfn.c_str(), m_type)
L95 of Certificate...

6) Question: is the certificate name taken from ServerConfig.xml?
If so, when is this done (haven't noticed this feature).

7) I see that the default certificate name in ServerConfig was changed from the past opcuacanopenserver to OpcUaServer. This of course is valid but IMO we should at least put information (in ChangeLog) about it, for people upgrading their projects to newer ServerConfig.xml. I see some pretty serious risk of merge errors here ...

8)  a detail: in main CMakeLists the copyright to Uni of Oviedo disappeared

9) I have no knowledge or testbed atm to check appveyor behaviour.

10) probably an improvement ticket for me: abstract out UA_Server (from open6) into OpcServer class (like in UA API); this would reduce the number of #ifdefs in BaseQuasarServer.cpp (which BTW seems the only file still burdened by dual backend support).

please comment further on my points

PR #18, opened 2016-08-30
OPCUA-653 reduce ifdefs

pnikiel:
The purpose is simplification and removal of backend-specific things in BaseQuasarServer class. Almost all backend-specific things were moved to OpcServer (existing previously) but now 2 implementations exist (one per each  backend).

This cleanup reduced number of ifdefs from 19 to 5 in cpp file, and also few in the header file.

The philosophy is similar to open62541-compat module: open62541-compatible implementation was provided for UA API OpcServer interface.

### DIFF
--- a/FrameworkInternals/original_files.txt
+++ b/FrameworkInternals/original_files.txt
@@ -175,6 +175,7 @@ File shutdown.h                             must_exist,must_be_versioned,md5=che
 File serverconfigxml.h                      deprecated
 File serverconfigxml_quasar.h               must_exist,must_be_versioned,md5=check,install=overwrite
 File opcserver.h                            must_exist,must_be_versioned,md5=check,install=overwrite
+File opcserver_open62541.h		    must_exist,must_be_versioned,md5=check,install=overwrite
 File version.h                              must_exist,must_be_versioned,md5=check,install=overwrite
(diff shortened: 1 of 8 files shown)


pnikiel:
I'd like to mention that a (successful) test with open62541 was performed but not yet with UA API so I'd be happy if someone could contribute ;-)

ben-farnham:
I see this error message when compiling on my SL6 environment, Unified Automation backend (v1.3.3). The toolchain file has this definition

> add_definitions( -DBACKEND_BACKEND_UATOOLKIT )

screenshot with build error:
capture

ben-farnham:
Note: Does compile if I change Server/CMakeLists.txt to omit src/opcserver_open62541.cpp

`add_library (Server OBJECT
	src/main.cpp 
	src/opcserver.cpp
	#src/opcserver_open62541.cpp OMITTED FOR TEST!
	src/shutdown.cpp 
	src/serverconfigxml.cpp 
	src/QuasarServerCallback.cpp
	src/BaseQuasarServer.cpp
	src/QuasarServer.cpp
)`

I guess the CMakeLists.txt needs to switch in either 
BACKEND_UATOOLKIT: src/opcserver.cpp
BACKEND_OPEN62541: src/opcserver_open62541.cpp

ben-farnham:
Fixed issue described above on branch.

PR #19, opened 2016-08-31
Windows compatibility changes

ben-farnham:
Hi All,

As you know, quasar windows support is a critical requirement for BE-ICS; we need quasar to produce various windows compatible servers. Now Damian's gone (with no immediate replacement in view) I'd like to ensure our selected windows CI environment, appveyor, tests each commit to master to ensure we don't deviate from windows compatibility. Or tell us when we do anyway.

Branch 'windows_compatiblity-changes' is currently building on appveyor. It was created by merging the latest master (31/08/2016 ~09h00) with windows/appveyor compatible changes.

Cheers,
Ben


pnikiel:
I've got few questions:
1) was this windows branch actually tested on SL6?
2) I don't understand the change in automatedBuild in Windows branch: it inhibits exception passage (no rethrow), so when a build step fails, the whole build process will continue instead of failing. IMO that's not so good.
3) The changes in configurationGenerators.py : what is the actual purpose? Why remaining functions from this module were kept using old approach and only generateConfiguration() uses new approach?
4) SET( CUSTOM_SERVER_MODULES open62541-compat) in open6_win_configuration: does this make sense? Since quasar 1.1.8 we have "optional modules" infrastructure so compat module should not be declared as custom server module anymore. It might be a relict from the past.
5) any particular reason for moving "CUSTOM_SERVER_MODULES" before inclusion of build_configuration? I guess it links with 4).
6) not an issue, just a comment: ** gitignore pattern is supported from 1.8 on (we have 1.7) but is not backwards-showstopper so It won't work but won't break anything either. So I'm OK with it.

ben-farnham:
1) ...tested on SL6?
Yep, on my SL6 dev machine. Works fine. In fact this branch builds on travis (linux) and appveyor (windows). Furthermore we must, as a dev team, strive to keep it that way.

2) ...automatedBuild in Windows branch...
I know, it's the dirtiest thing on this branch (so not _that_ bad). Here comes an excuse: On Appveyor, visual studio quietly barfs out some internal error, I truly have no idea what, and so fails the build. The strange this is that the server build itself actually completes fine though. I changed the appveyor script to test whether the server executable exists at then end. The build passes or fails depending on the presence of the executable.

3) ...changes in configurationGenerators.py : what is the actual purpose?...
We could just revert this I think, I found the debug output useful to diagnose CI build problems

4) ... CUSTOM_SERVER_MODULES open62541-compat...
5) ...CUSTOM_SERVER_MODULES...
I could not get optional modules to work with the open62541-compat. Lets look at this together, hopefully I was just doing something silly.

6) ...** gitignore pattern ... I'm OK with it.
Me too!

PR #23, opened 2016-10-31
Adding ByteStream to supported data types

pnikiel:
The feature was tested with a cache variable in both directions - reading and writing - from OPC-UA client.


pnikiel:
Thanks Ben!

PR #24, opened 2016-11-01
fix Certificate pt reference and constructor

meeludwig:
follow up of valgrind report: local var ref corected, contructor corrected. Please check and pull.


pnikiel:
Hello Michael,
was this tested?
Does valgrind no longer complain?

pnikiel:
Hello,
I'm closing this because it doesn't provide the solution. 
The tested solution is in #49

PR #28, opened 2017-05-29
added a switch to skip tedious merging

meeludwig:
Dear All,
I have extended the FrameworkInternals/deviceGenerators.py with a flag which allows to optionally skip the tedious kdiff:

overwriteProtection=1 
# 2=keep original (buffer A)
# 1=manual merge, as default
# 0=overwrite with generated (buffer B)

plus a small if-else py code extension.

We should probably adapt a more global config file to steer the python scripts, using a config file for that [and switching astyle behavior ?] Anyway, this flag can be hardcoded as well. Please take a look and pull it in if you like it. The behaviour stays as it is default==1. Set it to 2 for more fun.

Greetings, Michael

### DIFF
--- a/FrameworkInternals/deviceGenerators.py
+++ b/FrameworkInternals/deviceGenerators.py
@@ -24,6 +24,10 @@ from transformDesign import transformDesignVerbose
 from lxml import etree 
 from manage_files import get_list_classes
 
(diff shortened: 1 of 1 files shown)


pnikiel:
I think we discussed this one and chosen another approach in the end.

PR #31, opened 2017-06-05
Proposal for an easy selection of the ServerConfig.xml

pnikiel:
This feature enables to select "ServerConfig.xml" to load from the command line. 
The current situation (without the feature) is that ServerConfig.xml is "hardcoded" as server's executable path + "/ServerConfig.xml".

Please note that in order to avoid confusion, I propose to call this file "OPC-UA Settings File" rather than "configuration file" which IMO is reserved for description of instances.

The change is backwards compatible.

Here is what you get:
$ ./OpcUaServer --help
Allowed options:
  --config_file arg                     A path to the config file
*  --opcua_settings_file arg (=/tmp/opcua-579-test/bin/ServerConfig.xml)*
                                        A path to the OPC-UA settings file
  --create_certificate                  Create new certificate and exit
  --help                                Print help
  --version                             Print version and exit

Without arguments, server starts with standard settings:
$ ./OpcUaServer
***************************************************
 Server opened endpoints for following URLs:
     opc.tcp://pcatldcsp11.cern.ch:*4841*
***************************************************

With the argument, server starts with custom settings, e.g.:
$ ./OpcUaServer --opcua_settings_file ./ServerConfig-port12345.xml
***************************************************
 Server opened endpoints for following URLs:
     opc.tcp://pcatldcsp11.cern.ch:*12345*
***************************************************


pnikiel:
Just curious if anyone looked at this.

ben-farnham:
Hi Piotr,
Bit slow - just having a look now.
Cheers,
Ben

ben-farnham:
Changes look sane, I'll just take it for a drive on windows...

pnikiel:
Thanks! Appreciated!

PR #35, opened 2017-11-08
Fix for certificate creation for toolkit 1.5.5

pnikiel:
Apparently, in UA SDK 1.5.5, in order to just created PKI certificate repo, one has to do more than just loading the configuration, compared to UA SDK 1.3.3. 
This fix should address it; in addition I migrated few "cout" logs to LogIt.

Tested with UA SDK 1.3.3 and UA SDK 1.5.5.

Compare whitespace-insensitive:
https://github.com/quasar-team/quasar/pull/35/files?w=1

### DIFF
--- a/Server/src/opcserver.cpp
+++ b/Server/src/opcserver.cpp
@@ -277,32 +277,36 @@ int OpcServer::createCertificate ()
     // Check trace settings
     if (! d->m_pServerConfig->loadConfiguration().isGood() )
     {
-        LOG(Log::ERR) << "Couldn't open server's backend configuration";
+        LOG(Log::ERR) << "Couldn't open server's backend configuration -- check srvTrace logs";
         return -3;
     }
 
-    CoreModule* coreModule = new CoreModule;
-    if( coreModule->initialize() != 0)
+    CoreModule coreModule;
+    if( coreModule.initialize() != 0)
     {
-        LOG(Log::ERR) << "CoreModule::initialize failed.";
+        LOG(Log::ERR) << "CoreModule::initialize failed -- check srvTrace logs";
+        return -4;
     }
-    UaModule* uaModule = new UaModule;
+    UaModule uaModule;
     UaServer *pUaServer = 0;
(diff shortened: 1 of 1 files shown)


Server/src/opcserver.cpp
```diff
@@ -240,43 +240,71 @@ int OpcServer::setCallback(OpcServerCallback* pOpcServerCallback)
...
+        LOG(Log::ERR) << "CoreModule::startUp failed.";
+    }
+    if( uaModule->startUp(coreModule) != 0)
+    {
+        LOG(Log::ERR) << "UaModule::startUp failed.";
+    }
+
+    uaModule->shutDown();
+    coreModule->shutDown();
+
```
ben-farnham: Hi Piotr,
Do we need to allocate uaModule & coreModule on the heap (lines 284, 289)? Just thinking that if, say, line 291 threw an exception for some reason then we leak (no?). Also, do the shutdown() methods of uaModule & coreModule destruct the objects?

As a suggestion: would it work if uaModule and coreModule were stack based variables?

Cheers,
Ben

pnikiel:
That's an entirely fair comment! Thank you. I will implement it.

(In the mode, the server will quit just after creating the certificate,
so the risk of 'operational' memory leak is none - but let's do good
work no matter where.)

best, piotr

pnikiel:
OK, implemented. Funny that nobody noticed missing returns() after errors were to be printed ;-) Now fixed as well.

pnikiel:
Hi, 

if the Configuration.cpp does not compile, to me it sounds like a bug.
In such a case you should file a ticket and attach the corresponding
Design.

On another note, indeed there is nothing like Parent when given class
is a children of many different quasar classes, and it is intentional.
One can discuss things there, but certainly it won't be done quickly.
On contrary, it is rather easy to get around this if instead of looking
"up" (toward parents) you look down (from parents to children), this
nuisance completely disappears then.

cheers -- p

PR #40, opened 2018-03-27
Working version with support for --project_binary_dir

pnikiel:


ben-farnham:
I guess there should be some documentation added to the help for the 'build' command to instruct users that --project-binary-dir is an option and how to use it.

image

PR #46, opened 2018-05-04
Quasar Thread Pool, and its application to source variables

pnikiel:
There is an issue with UASDK ThreadPool: it has a fixed maximum task/thread factor of 4. This prevents some classes of applications, e.g. where a very big number of requests have to be accepted at one given moment. 

After a brief consideration of boost::asio threadpool, I rolled out this one. It is past initial tests where I managed to submit 50000 requests distributed into varied number of threads. Seems fine so far but will be additionally tested, of course.

What's missing: the maximum number of requests of source variables is not exposed in the config yet. So far hardcoded. Requires some changes of Meta.

### DIFF
--- a/Common/include/QuasarThreadPool.h
+++ b/Common/include/QuasarThreadPool.h
@@ -1,0 +1,55 @@
+/*
+ * QuasarThreadPool.h
+ *
+ *  Created on: 4 May 2018
+ *      Author: pnikiel
+ */
+
+#ifndef COMMON_INCLUDE_QUASARTHREADPOOL_H_
+#define COMMON_INCLUDE_QUASARTHREADPOOL_H_
+
+#include <mutex>
+#include <vector>
+#include <thread>
+#include <list>
+#include <condition_variable>
+
+#include <statuscode.h>
+
+namespace Quasar
+{
+
+class ThreadPoolJob
+{
+public:
+    virtual ~ThreadPoolJob() {};
+
+    virtual void execute() = 0;
+};
+
+class ThreadPool
+{
+public:
+    ThreadPool (unsigned int maxThreads, unsigned int maxJobs);
+    ~ThreadPool ();
+
+    UaStatus addJob (ThreadPoolJob* job);
+
+private:
+    void work();
+
+    std::mutex m_accessLock;
+    bool m_quit;
+    std::vector<std::thread> m_workers;
+    std::list<ThreadPoolJob*> m_pendingJobs;
+
+    const unsigned int m_maxJobs;
+
+    // this is the notification business for conditional variable notification
+    std::condition_variable m_conditionVariable;
+
+};
+
+}
+
+#endif /* COMMON_INCLUDE_QUASARTHREADPOOL_H_ */
--- a/Common/include/QuasarThreadPool.h
+++ b/Common/include/QuasarThreadPool.h
@@ -1,8 +1,22 @@
-/*
+/* © Copyright CERN, 2018.  All rights not expressly granted are reserved.
  * QuasarThreadPool.h
  *
  *  Created on: 4 May 2018
- *      Author: pnikiel
+ *      Author: Piotr Nikiel <piotr@nikiel.info>
+ *
+ *  This file is part of Quasar.
+ *
+ *  Quasar is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Lesser General Public Licence as published by
+ *  the Free Software Foundation, either version 3 of the Licence.
+ *
+ *  Quasar is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


ben-farnham:
Am in the office again now. I'll take a look...

Common/src/QuasarThreadPool.cpp
```diff
@@ -0,0 +1,96 @@
...
+    for (unsigned int i=0; i<maxThreads; ++i)
+        m_workers.emplace_back( [this](){this->work();} );
+}
+
+ThreadPool::~ThreadPool ()
+{
+    LOG(Log::TRC) << "Stopping threadpool";
+    m_quit = true;
+    m_conditionVariable.notify_all();
+    std::for_each( m_workers.begin(), m_workers.end(), [](std::thread &t){t.join();} );
```
ben-farnham: A timeout on the join? That way if some jobs do not shut down properly then at least we know which ones they are (if we have some way of ID-ing the jobs as per previous toString()/onError() comment).

Common/include/QuasarThreadPool.h
```diff
@@ -0,0 +1,69 @@
...
+
+    UaStatus addJob (ThreadPoolJob* job);
+
+private:
+    void work();
+
+    std::mutex m_accessLock;
+    bool m_quit;
+    std::vector<std::thread> m_workers;
+    std::list<ThreadPoolJob*> m_pendingJobs;
```
ben-farnham: or maybe a std::queue

pnikiel: Maybe ;-) But I don't see a strong technical advantage here. I will consider it though.

Common/src/QuasarThreadPool.cpp
pnikiel: I have double thoughts here because of:
1) it is never clear what is the right timeout (I wouldn't like to add this to configuration unless necessary),
2) killing threads brutally might cause more mess (e.g. segfauls, core dumps) than being patient and letting the end user do that using "kill" tool if really necessary.

Common/src/QuasarThreadPool.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+UaStatus ThreadPool::addJob (ThreadPoolJob* job)
+{
+    {
+        std::lock_guard<std::mutex>lock (m_accessLock);
+        if (m_pendingJobs.size() >= m_maxJobs)
+        {
+            LOG(Log::ERR) << "The threadpool is already full, cant add new jobs. Enlarge the threadpool";
```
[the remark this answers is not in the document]
pnikiel: Will do.

Common/src/QuasarThreadPool.cpp
```diff
@@ -0,0 +1,96 @@
...
+            unsigned int size = m_pendingJobs.size();
+            lock.unlock();
+            LOG(Log::TRC) << "Removed job from the threadpool, current number of jobs is:" << size;
+            try
+            {
+                job->execute();
+            }
+            catch (...)
+            {
+                LOG(Log::ERR) << "Job has thrown an unhandled exception.";
```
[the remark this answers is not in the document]
pnikiel: Sounds like a nice extension. Thanks.

Common/include/QuasarThreadPool.h
ben-farnham: As you wish: it is a cosmetic/subjective point. I mentioned it because I thought queue seemed a clearer indication of the intent of the code than list

Common/src/QuasarThreadPool.cpp
ben-farnham: Yep, I always seem to end up at roughly the same point when dealing with threads: How to handle the badly behaved instances that don't stop when they're supposed to? How long to wait etc.

To add some precision to my statement... I hadn't actually meant we should do anything brutal to lingering threads, rather that we should just log the threads that take a long time to stop - the sort of information that could help a developer improve their product. 

How long is a long time? Should it be configurable? I'd say if a thread hasn't stopped in 5s that's a long time: worth a warn/info entry in the logfile anyway.

Common/include/QuasarThreadPool.h
pnikiel: OK I bought your idea.

Common/src/QuasarThreadPool.cpp
pnikiel: I like your idea. But it seems std::thread is a bit behind boost::thread in the sense that it doesn't yet support timed_join. I will try to see what can be done with std::thread::joinable().

pnikiel: Hi @ben-farnham, I failed with std::thread. 
One could do with boost::thread but then better to go for thread_group because timed_join makes sense only on the whole group (to execute join concurrently rather than one after another).
So what I will merge is sth that prints a message in INF log level that the threadpool is being stopped and that it might require time, and a corresponding message that it's stopped.

There clearly is some room for user satisfaction improvement, so let's keep your idea this way.

PR #50, opened 2018-06-27
Reinstate windows continuous integration build (appveyor) with better package management

ben-farnham:
Hi All,

This pull request reinstates the windows continuous integration build. This is (I believe) a better approach to the CI build than we had previously. The previous approach contained a lot of hard-coded paths, URLs and weird blobs squirrelled away in CERN drop-boxes etc. Now the approach to provisioning the windows CI build machine is built on the emerging windows package manager tools chocolately (apps) and nuget (libs). A more solid way of doing things I think.

A special request: Can this jump the pull request queue (i.e. be the next pull request merged to master). And what makes this _**so*_* special you ask? I'd like to ensure that the current list of queued pull requests do not break the windows CI build.

Once master has two green build badges - travis (linux) and appveyor (windows) - then we easily gauge whether other pending pull requests will break the build on one, or more platforms.

Cheers,
Ben


pnikiel:
Hello @ben-farnham,

before some more detailed comments:
1. Please squash this branch. This is currently 85 commits out of which most is trial and error, and a simple merge will append all of that to quasar master history. I'd like to ask you (and everyone around) that such trial and error commits come as one final commit. Please let me know if you like some support in squashing it (the process itself is a quick one, don't worry).
2. Please, if possible, append relevant JIRA ticket either in the branch title or somewhere in the description. As you know we maintain Quasar Changelog which is generated from JIRA and right now I do not know how this improvements correlates with Quasar JIRA.

FrameworkInternals/optionalModules.py
```diff
@@ -172,65 +172,64 @@ def enableModule(moduleName, tag="master", serverString=""):
 	if not _checkTagExists(moduleInfo[moduleName]["url"], tag): return
 
 	baseDirectory = os.getcwd()
-	fwInternalsDir = baseDirectory + os.path.sep + "FrameworkInternals"
-	os.chdir(fwInternalsDir)
```
pnikiel: I've got doubts if removing os.chdir here is a good idea, seeing what happens in the next lines. Please clarify.

FrameworkInternals/optionalModules.py
```diff
@@ -172,65 +172,64 @@ def enableModule(moduleName, tag="master", serverString=""):
...
 
 	baseDirectory = os.getcwd()
-	fwInternalsDir = baseDirectory + os.path.sep + "FrameworkInternals"
-	os.chdir(fwInternalsDir)
+	fwInternalsDir = os.path.join(baseDirectory, "FrameworkInternals")
+        enabledModulesDir = os.path.join(fwInternalsDir, "EnabledModules")
 
 	# Check first if module is maybe already present with different version
 	#
-	tagFileName = "EnabledModules/"+moduleName+".tag"
```
pnikiel: Why renaming this variable? 
It stores the name of the tag file so the original name seems far more appropriate.

pnikiel:
So BTW what is the ETA for this? As discussed yesterday I'm OK with this going first as long as it doesn't postpone the other things significantly (and some of the other things are already pending review for 4 months).

ben-farnham:
OK, addressed both comments regarding optionalModules.py by reverting to (very nearly) the current master version. Only change is whitespace and replacing a linux 'pwd' with a python platform non-specific equivalent.

Will squash and merge.

PR #51, opened 2018-07-12
OPCUA-1036 isKey doesn't work without device logic and with arrays

pnikiel:
Implemented in the design validator.


ben-farnham:
Yep, looks good: OK to merge I'd say.

PR #52, opened 2018-07-12
OPCUA-1061 rename configurationhxx target

pnikiel:
The reason for such a change is that the target was named exactly as the file it generates (which was not the most optimal decision), and that confused "ninja" builder.


ben-farnham:
Hi Piotr, in order to give this PR a green light as regards CI builds I think the feature branch just needs an up-merge of master applied to it. I notice that appveyor.yml in this branch is missing the 'python pip' update that resolved the windows CI builds this morning.

Alternatively you could just merge to master. The build status panel reports that this PR (i.e. this feature branch auto-merged with master) will pass both travis and appveyor CI builds.

PR #53, opened 2018-07-19
Implemented and tested

pnikiel:
The reason of exposing device logic mutex (when it is declared in the Design of course) is to enable RAII-style constructs like:
std::lock_guard<decltype( getLock())> lock (getLock());

### DIFF
--- a/Device/designToDeviceBaseHeader.xslt
+++ b/Device/designToDeviceBaseHeader.xslt
@@ -115,6 +115,7 @@
 		<xsl:if test="fnc:classDeviceLogicHasMutex(/,$className)='true'">
 		void lock () { m_lock.lock(); }
 		void unlock () { m_lock.unlock(); }
+        boost::mutex &amp; getLock () { return m_lock; }
 		</xsl:if>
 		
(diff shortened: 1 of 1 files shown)


ben-farnham:
Looks good to me. 

One thought - I remember we agreed C++11 was OK these days, how about we migrate away from boost threading stuff. No need to search and destroy everything related to boost threading in quasar but perhaps it makes sense to only expose C++11 in this case?

Exposing boost::mutex forces the developer to include boost headers, use the boost API etc. It might be a slightly 'cleaner' user experience if we expose a std::mutex.

What do you think?

pnikiel:
Hi Ben,

std::mutex also requires to include a header file; which in both cases will be included by "this" include file (i.e. the one generated from this XSLT).

I do agree with your points in general but my experience so far with std::threading is that (in C++11, that is, and gcc4.x line that is agreed) it serves only basic needs. For instance, while I was designing Quasar thread pool, it turned out that std::threading was simply missing too many things like timed joins, etc etc.

In other words, migrating from boost threading to std threading could be backwards incompatible because of missing API.

PR #57, opened 2018-07-26
Minimizing redundancy and code-bloat of Python scripts, maximizing managability

pnikiel:
Improvements:
1. Reduction of redundancy and copy-paste code in multiple Python functions,
2. Moving of (copy-pasted based) transform invocation to a table of transform with one entry point (transformByKey)
3. Reworking exception system into frameworks exception (which leave full stack-trace and can be inspected by PDF) and into external tool exceptions (which won't leave full stack-trace as that would not help)


ben-farnham:
Hi Piotr,

Will review this afternoon (mainly in meetings this morning, sadly)

Cheers,
Ben

FrameworkInternals/generateHonkyTonk.py
```diff
@@ -1,30 +0,0 @@
...
-THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
-@contact:    quasar-developers@cern.ch
-'''
-
-import os
-from transformDesign import transformDesignVerbose
-
-def generateHonkyTonk():
-	"""Generates honkyTonky.cc, a special class intended for testing your namespace with simulated (random) values."""
```
ben-farnham: have we lost this documentation as to what the command
quasar.py generate honkyTonk MY_CLASS
actually does?

I feel end users probably genuinely would profit from having some help as to what this command does.

FrameworkInternals/transformDesign.py
```diff
@@ -27,77 +27,146 @@
...
+    KEY = 0
+    XSLT_PATH = 1
+    OUT_PATH = 2 
+    SOURCE_OR_BINARY = 3   
+    CPP_FORMAT = 4
+    OVERWRITE_PROT = 5
+    ADDITIONAL_PARAM = 6
+    
+
+QuasarTransforms = [
```
ben-farnham: On a general note I very much like this table and the corresponding simplification it allows in the code, it's a nice idea and implementation

FrameworkInternals/generateHonkyTonk.py
pnikiel: @ben-farnham,
my proposal to solve it is the following: when an user calls any command with -h, a web browser would be opened, positioned on an anchor (a paragraph) about particular command.

This can be done using Python's standard module:
import webbrowser
webbrowser.open('http://google.com')

advantages:
* would let reuse the same clean python logic,
* the documentation could be of better quality (one can embed figures, tables in HTML instead of few lines of text in python's doc) and it could be more detailed

I could make an initial quasarCommands.html copy-pasting the existing documentation and aim for refinement as we go. Makes sense?

FrameworkInternals/transformDesign.py
```diff
@@ -27,77 +27,146 @@
...
+    
+# column IDs
+@enum.unique
+class FieldIds(enum.Enum):
+    KEY = 0
+    XSLT_PATH = 1
+    OUT_PATH = 2 
+    SOURCE_OR_BINARY = 3   
+    CPP_FORMAT = 4
+    OVERWRITE_PROT = 5
```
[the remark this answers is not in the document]
pnikiel: corrected now.

FrameworkInternals/generateHonkyTonk.py
ben-farnham: Yep, makes sense.

pnikiel: Thanks. It will be done under another ticket:
https://its.cern.ch/jira/browse/OPCUA-1111
aimed to enter the same quasar release.
I will merge this PR today when the final CI check is finished.

PR #60, opened 2018-08-01
Moved old docs into quasarCommands.html and made an interface to it v…

pnikiel:
…ia python webbrowser

Note that the documentation I attached is taken from the old docs attached to particular functions and apart from few cases leaves A LOT to be desired.

@ben-farnham would you have a chance to test under Windows if e.g. the following brings up the docs?

./quasar.py create_project -h


ben-farnham:
Hmm fails. Missing python import (I guess then this means there's something missing from python_requirements.txt) 

I tried the following (which used to give the complete command list), now gives missing import for 'enum' error (python version 2.7.14)
image

Similarly for _create_project_
image

ben-farnham:
Aha, yep - python pip installed enum34 and is OK.

ben-farnham:
Also just tried 
{code}
python quasar.py generate honkytonk --help
{code}

It _does_ open the HTML help, but not at the honkytonk anchor - just opens at top of HTML page.

pnikiel:
@ben-farnham ,
what is the URL visible in the address bar when you did 

python quasar.py generate honkytonk --help

?

ben-farnham:
image

pnikiel:
Hi @ben-farnham ,
I suppose this relates to file:// URI handling in Windows. 
I do not know why this standard Python's module is not passing this correctly.

It deserves a ticket but it's rather a small flaw in my opinion, and it doesn't cut users from the help page.

My proposal to move on is the following:
- merge this,
- open a separate ticket,
- let BE/ICS figure out how this particular anchor should be passed in Windows

OK?

ben-farnham:
OK, I suppose... It _is_ less than good that quasar on windows has to play catch up, feels a bit a 2nd class citizen and I'm wary of that. However - as you say - the functionality is not critical, so we let this one go.

As regards the enum34 (missing import error); I think that merits an entry in python_requirements.txt. 

Add that entry and I'd say ready for merge.

pnikiel:
Hi Ben,
OK, the enum34 has been fixed so I'm merging this.
Please note that I can't take care of Windows compatibility myself, It's not my intent for it to be a 2nd class citizen but since it's been brought up by BE/ICS, I'd expect some BE/ICS involvement in figuring out how to make **standard** things behave fine. (please note, I did use standard Python module to open a webpage, perhaps there is something wrong in the usage mode, and that is where I expect some help).

Thanks for the review!

PR #61, opened 2018-08-01
ported argument passing logic into the new python scripting

pnikiel:

PR #62, opened 2018-08-06
Merge-back the out-of-source

pnikiel:
This is the full-featured out-of-source build capability.
The last PR #61 equipped individual quasar generators to understand --project_binary_dir, this PR benefits from it via update CMake statements.

The end-user advantage is that they can do:

./quasar build --project_binary_dir   <somewhere>

and given quasar server should be build in <somewhere> without polluting the source directory.

### DIFF
--- a/AddressSpace/CMakeLists.txt
+++ b/AddressSpace/CMakeLists.txt
@@ -16,19 +16,19 @@
 
 # Author: Piotr Nikiel <piotr@nikiel.info>
 
-include(cmake_generated.cmake)
+include(${PROJECT_BINARY_DIR}/AddressSpace/cmake_generated.cmake)
 
-	add_custom_command(OUTPUT include/ASInformationModel.h src/ASInformationModel.cpp
-	WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-	COMMAND ${PYTHON_COMMAND} quasar.py generate info_model
-	DEPENDS ${DESIGN_FILE} ${PROJECT_SOURCE_DIR}/AddressSpace/designToInformationModelHeader.xslt ${DESIGN_FILE} ${PROJECT_SOURCE_DIR}/AddressSpace/designToInformationModelBody.xslt 
-	)
-	
-	add_custom_command(OUTPUT src/SourceVariables.cpp include/SourceVariables.h
-	WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-	COMMAND ${PYTHON_COMMAND} quasar.py generate source_variables
-	DEPENDS ${DESIGN_FILE} ${PROJECT_SOURCE_DIR}/AddressSpace/designToSourceVariablesBody.xslt  ${PROJECT_SOURCE_DIR}/AddressSpace/designToSourceVariablesHeader.xslt DeviceBase
-	)
+add_custom_command(OUTPUT ${PROJECT_BINARY_DIR}/AddressSpace/include/ASInformationModel.h ${PROJECT_BINARY_DIR}/AddressSpace/src/ASInformationModel.cpp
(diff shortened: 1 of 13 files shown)


pnikiel:
@ben-farnham as a supporting evidence: just tried this with our most complex OPC-UA server so far (SCA) which has some optional modules and stuff like this and it built just fine out-of-source.

One thing to remember is to make sure that switching to out-of-source build from in-source requires clearing CMakeCache.txt, I will create a ticket for this, otherwise CMake gets very much confused.

pnikiel:
Hi @ben-farnham ,
I do see your point and it does make some sense. However my argumentation would be that the optional modules get fetched and configured in the build process - and that means that the build process would pollute the source directory which is what we want to avoid. Furthermore you could not run two or more concurrent out-of-source builds if your idea was applied, for exactly same reason (that is, modification of source dir during the build).

I'd say therefore that if someone uses a custom mechanism to deploy their modules (e.g. "git submodule") then they are free to deploy them in source, as you state, but if someone uses our optional modules, the binary dir seems more appropriate to me.

Perhaps we should adopt a more general definition of "binary dir" meaning "everything that gets created in the build process" (where created is one of: fetched, built, generated etc).

How would you relate to that?

ben-farnham:
A couple of opinions from me

* I like the simplicity of handling source code and generated build stuff in distinct locations. It's a clean concept and provides a tidy way to work. I believe this is the basis of cmake's CMAKE_BINARY_DIR notion, do we want to add our own specific flavours to that notion? (my thoughts on this: _nope_).

* I'd find it weird that some source code lives here but other source code lives there - weird is trouble, especially for new users (and old users with poor memories, e.g. me). Optional modules contain source code; I'd like that source code to live with other source code.

* Quote "_optional modules get fetched and configured in the build process_". Are we talking about _prepare_build_? This command confuses me; it's the twilight zone. To fetch code and configure code is code-centric. Perhaps we should untangle the build-centric part of prepare_build? If so that sounds like a deeper conversation than textual exchanges allow for.

* A use case: I'm writing a server for centos7 and rasp-pi. This server uses a new optional module. As I develop my code I cross compile for the 2 target platforms, so I can occasionally deploy and test. To me, I'd sacrifice concurrent builds for the convenience of the optional module code being in one place.

pnikiel:
Hi @ben-farnham ,

OK for me to rethink optional modules and prepare_build but not right now and not for this quasar release. In fact the points you raised right now (e.g. about code-centricity) were exactly the same points I raised back then in 2016. It's a pity you're bringing this up with 2 years delay rather than then when there were some real discussions and testing and reviews between Stefan and me.

For your point about where source code goes... it's a vague point. Most of generation output is C++ source code and it does go to binary dir and we see no problem with that. So yes we have some twilight zone here.

My proposal is that would you please get prepared for today's meeting and you demonstrate your use cases listing pros and cons for deployment path of optional modules and we will take the decision today.

ben-farnham:
"_Most of generation output is C++ source code and it does go to binary dir and we see no problem with that_"

If we're talking about generated AddressSpace C++ source code etc - I'd say that's different: We do not expect (nor want) quasar developers to put their fingers in it: it's some kind of 'quasar compiler output' that just happens to be C++ source code.

Optional module C++ code is different.

pnikiel:
@ben-farnham you can always attach an optional module to your source tree using "git submodule" or even plain "git clone" and add one line into ProjectSettings.cmake. The result will be *identical* to what you are talking about. The VERY SAME optional module repo you can later on include as a normal optional module without any changes in it.

If you still insist on deploying optional modules at the build phase into source tree, please be ready for discussion in the afternoon.

pnikiel:
Just for the record, we concluded to spawn the "optional modules optimization" as a separate ticket:
https://its.cern.ch/jira/browse/OPCUA-1118

PR #63, opened 2018-08-07
Some initial working version with xs:enumeration and xs:pattenr for c…

pnikiel:
…onfigentry

More detailed discussion:
https://its.cern.ch/jira/browse/OPCUA-856

### DIFF
--- a/Configuration/designToConfigurationXSD.xslt
+++ b/Configuration/designToConfigurationXSD.xslt
@@ -51,6 +51,35 @@ xsi:schemaLocation="http://www.w3.org/1999/XSL/Transform schema-for-xslt20.xsd "
 	
 </xsl:function>
 
+<xsl:template name="configRestrictionForScalar">
+<xsl:choose>
+    <xsl:when test="not(d:configRestriction)">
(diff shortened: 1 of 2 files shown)


pnikiel:
@ben-farnham a context information for your review: please note that this is a feature demo and not a full implementation, e.g. it does aim for configentry restrictions but not cachevariables (note the mechanism is very similary in the end). Also it doesn't aim for arrays value restrictions (yet).

Once we converge that this is the right concept (also from the end-user point of view) I will provide a more complete PR.

ben-farnham:
Generally I like it; the syntax required in Design.XML forced out by the XSD changes make sense and results in a design that, to my mind, is clear and readable.

ben-farnham:
So this PR is not a 'merge to master PR', yet - it's rather a preview of the shape of things to come?

pnikiel:
It does work and you can try it out, but it works only for config entries (which would probably make 80% or so of intended use cases). Since you said that Design-wise it seems satisfactory, I will expand the scope to also cover cachevariables and I will see if I can quickly cover arrays.

pnikiel:
OK this one supports pattern- and enumeration-based XSD restrictions for config-entries and cache-variables no matter if they are scalars or arrays. Let me know if there are any extra comments. Ready for merge from my point of view.

PR #65, opened 2018-08-30
Fix for generate device --all

pnikiel:


FrameworkInternals/deviceGenerators.py
```diff
@@ -39,9 +39,8 @@ def generateDeviceClass(context, *classList):
 	for aClass in classList:
 		generateOneDeviceClass(context, aClass)
 	
-def generateAllDevices():
+def generateAllDevices(context):
```
ben-farnham: Feel free to ignore this comment - it's highly subjective, probably a bit rant-y.

I've always found the variable name 'context' frustratingly opaque when trying to understand code (and, since I'm writing about annoying variable names, 'data' is another one). I just mean that on reading code, variables named context doesn't reveal much about what is contained within, or is what it's used for. I mean all code executes in some context (and deals with data for that matter). So the reader either has to search around up the call stack to see where context came from (or downwards, to see how/where it's consumed. Or both).

If you leave as is there will be no further noise from me on the subject. It's not a big deal.

pnikiel:
Hey,

yes, why not to improve it? 
The term "context" in quasar python scripts always relates to the set of the following variables:
* project source dir,
* project binary dir,

In future it might be extended with "build type" etc... common things that must be passed to the build system.

Could you please suggest a more descriptive name? If you do so, I will create a separate improvement ticket.

ben-farnham:
So, current usage (source/binary dir) sounds like we could exchange variable names:
context -> targetDirectory.

I think a variable called 'targetDirectory' as a parameter to a function called generateXXX should be fairly easy to comprehend by the casual code reader.

In future, if we opt to send more information to generateXXX then I guess we can choose to
1. add new (aptly named) parameters, e.g. 'buildType'
2. expand some existing parameter by adding the new information to it, then rename that parameter accordingly with whatever we added. Maybe 'generatorSettings'?

pnikiel:
I think we have to limit ourselves to passing a Python dictionary with different fields (currently those two - projectSourceDir, projectBinaryDir and in near future maybe buildType). There are lots of reasons not to use individual arguments (especially that we converged to using a dict a month ago or so).

Therefore what we can do is to rename this dictionary from "context" to something else if there is an idea for it.

So, is there an idea for such a variable rename? ;-)

ben-farnham:
OK then - how does 'generatorSettings' sound?

pnikiel:
It sounds too thin to express what it does. E.g. projectSourceDir (a child field) is not really a generator setting, it is a property of given build. Maybe we could do "buildProperties" or "buildContext".

ben-farnham:
I like 'buildProperties' - 'buildSettings' also OK. I'd say either of those.

pnikiel:
OK, pls create a ticket and assign to me.
Merging this one for now.

PR #66, opened 2018-08-30
Opcua 743 obtain logit src from the one true repo

ben-farnham:
Hi All,

PR as discussed in our Tuesday (Aug 28th) meeting regarding keeping the LogIt source code in one single place.

This basically incorporates 2 large-ish changes:

1. LogIt source code has gone from the quasar repo. Now, quasar should be cloned with the --recursive flag (documented in quasar's README.md) to gather up LogIt source from its dedicated repo. Projects created from your quasar clone will have their own (dumb) copy of LogIt - this is so you can work unaffected in 'aeroplane mode'. Since LogIt source code is gathered in at clone time you can also create new projects with your network cable out.

2. The LogIt changed slightly (for ticket [OPCUA-743](https://its.cern.ch/jira/browse/OPCUA-743)). Now log messages can be 
a) Channeled through a component via a component handle - this is largely analogous to how LogIt has worked since its inception.
b) Channeled through a component via a component name (a string). This is largely to satisfy the requirement that disparate logging sources do not need to share LogIt component Ids.

This change *will* require a change to existing quasar servers as regards initialising logging components. it's a big change though, I don't think so anyway, and I made a serious attempt at documenting how it should be done in file Documentation/LogIt.html

Cheers,
Ben


pnikiel:
Is it ready for review or shall we wait?

Does it use the same LogIt that was reviewed by me (i.e. from the standalone LogIt repo as of Dec 2017 or so)?

I'd like to ask that you have a quasar JIRA ticket for this and that you specify (preferably in the ticket) which quasar release it should enter.

I'd also like to ask you that if you wish to propagate changes in LogIt into quasar (i.e. change the LogIt commit-id) it is treated as a normal review (equally to how it would look like in case LogIt was a deep-copy in quasar).

ben-farnham:
Ready for review.

As regards tickets - https://its.cern.ch/jira/browse/OPCUA-743? Or you mean something else/extra?

Target quasar version - hmm, not sure. Howabout the team looks at the documentation regarding the changes needed to initialise LogIt now (vs the olden days) and we chat about that - since versioning increments should be aligned with backwards compatibility etc.

Propagating LogIt changes into quasar - yep, I agree. This is the first review in that series.

pnikiel:
Hi Ben,

so tried to just see if it works and the only way to get this to a quasar working repo was by cloning a new repo in this bizzare way:

git clone https://github.com/quasar-team/quasar.git quasar4 --recursive -b OPCUA-743-obtain-logit-src-from-the-one-true-repo

all other methods (and in principle with "git submodules update") just leave LogIt directory empty.

Please fix it first.

I don't want to get my quasar git repo killed (i.e. recreated from a fresh clone) just because of that (esp. that I have got plenty of private branches with different quasar features). There's something wrong.

pnikiel:
As for the JIRA ticket:
https://its.cern.ch/jira/browse/OPCUA-743
it's titled: "LogIt: meet requirement to include optional components without knowledge about LogIds" so completely unrelated about ditching out LogIt deep-copy from quasar and instantiating from submodules. Please create a ticket that documents your actual change in quasar.

ben-farnham:
Hmm, thanks (re: submodule) I'll take a look. I must say so far I'm not loving git submodule functionality. It feels like trying to walk through a minefield wearing snow shoes.

pnikiel:
I think we might pick another solution and it is the right moment to evaluate them cheaply right now.
Of course as long as we maintain backwards compatibility (to a sensible level) and as long as they have favourable properties.

ben-farnham:
OK, cannot recreate, neither in windows nor linux. I'm using git v2.18.0 (windows) and v1.8.3.1 (CC7). Here's what I tried

1. cloned quasar repo (master branch)
2. checked out this branch
3. git submodule update --init --recursive

Could you try those commands and see what you get?

pnikiel:
OK point taken, "git submodule update --init --recursive" did the job. For some reason just "update" didn't work but I assume it's been my fault. Now moving on to the factual tests and review.

ben-farnham:
Maybe I should add a line to the quasar README.md; Users with existing quasar repos should update with
```
git submodule update --init --recursive
```
Agree?

AddressSpace/include/ASNodeQueries.h
```diff
@@ -148,7 +148,7 @@ namespace AddressSpace
         {
             unsigned int nMatched=0;
 #ifdef BACKEND_OPEN62541
-      BOOST_FOREACH( const UaNode::ReferencedTarget &target, *(startNode->referencedTargets()) )
+    for(const UaNode::ReferencedTarget &target : *(startNode->referencedTargets()) )
```
pnikiel: not part of the feature i suppose. please clean up.

Meta/CMakeLists.txt
```diff
@@ -16,6 +16,12 @@
 
 # Author: Benjamin Farnham <benjamin.farnham@cern.ch>
 
+project(Meta CXX)
+cmake_minimum_required(VERSION 2.8)
+
+set (CMAKE_CXX_STANDARD 11) # works for CMake v3.1+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11") #for older CMake versions
```
pnikiel: same as above

pnikiel:
Could you fix meta.cpp:265 at the same time?
 abort(); // TODO: Ben: pass the information back in more human way, perhaps?

just been struck by it while testing.
I'd suggest throwing a normal std::runtime_error or alike.

ben-farnham:
Yep - will do.

pnikiel:
Apart from the above, the test with a quasar server that doesn't do component LogIt passed fine. 
Do I need any specific instructions to upgrade a server with specific components?

ben-farnham:
Yep - you should find the documentation you need in Documentation/LogIt.html.

It doesn't specifically mention upgrading from an 'old' server, since the practice of storing LogIt component IDs was never a mandatory thing.

Rather it details the suggested practice for registering LogIt component Ids (using LogIt v0.1.0).

If you don't find what you need in that document then let me know where/how you're getting stuck and I'll update the docs accordingly.

AddressSpace/include/ASNodeQueries.h
ben-farnham: Weirdly it is - I removed BOOST_FOREACH from Meta, which depends on AddressSpace. Without this change ASNodeQueries fails to compile. Weirder still, works on linux but not on windows.

So, I changed to use C++11 range based for. Result: happy in both platforms.

pnikiel: I can conditionally accept this change, I suppose the problem is that simply ASNodeQueries.h doesn't  include boost_foreach.hpp which it should.
For future I'd strongly prefer to avoid this type of entangled fixes though.

ben-farnham: OK. The choice was
a) add include for boost
b) replace boost with std C++11

Went (b)

Meta/CMakeLists.txt
```diff
@@ -16,6 +16,12 @@
 
 # Author: Benjamin Farnham <benjamin.farnham@cern.ch>
 
+project(Meta CXX)
+cmake_minimum_required(VERSION 2.8)
+
+set (CMAKE_CXX_STANDARD 11) # works for CMake v3.1+
```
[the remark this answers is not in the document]
ben-farnham: Done

pnikiel:
okay so now tried with ATLAS Wiener Server which does include two components - Device and CanInterface.

Present state:
* component ids are "statically" declared in LogItComponentIds.h,
* in QuasarServer.cpp there is:
Log::initializeLogging(Log::INF, {ComponentAttributes(LOG_DEVICE, "Device"), ComponentAttributes(LOG_CAN_INTERFACE, "CanInterface",Log::INF)});

I guess I should:
1. invoke Log::registerLoggingComponent per component to add those two,
2. replace entries from LogItComponentIds.h by one-time call to resolve the handle at startup.

correct?

ben-farnham:
Yep, sounds correct to me.

ben-farnham:
And after registration, you should be able to log by either
a) the component handle
b) the component name (as used at registration)

with (a) being the preferred -faster- option

pnikiel:
OK tested it and worked fine.
Since I already had a lot of code with handles (which used to be just enums but doesnt matter then I did):
// global scope
Log::LogComponentHandle LOG_DEVICE;
Log::LogComponentHandle LOG_CAN_INTERFACE;

// in init
    LOG_DEVICE = Log::registerLoggingComponent("Device");
    LOG_CAN_INTERFACE = Log::registerLoggingComponent("CanInterface");

and essentially porting was done, behaviour seems to be correct.

pnikiel:
I'm OK with merging this to quasar 1.2.7. But I insist on having a JIRA ticket (component quasar) that says that LogIt is now taken from its native repo rather than quasar deep copy (the existing ticket 743 seems to relate to LogIt itself rather than to changes in quasar).

ben-farnham:
Good to know it works in someone else's environment too. Appreciate the review effort - thanks.

So, will...
1. make the JIRA ticket (with target release 1.2.7)
2. merge to master (squash and merge)

sound ok?

pnikiel:
I don't think that squash is necessary in this case.

Apart from that the rest sounds great, kudos!

PR #73, opened 2018-10-15
Implemented OPCUA-1109

pnikiel:
If --project_binary_dir is not specified, it now defaults to "build" subdir of the project.


pnikiel:
Hi @ben-farnham ,
I'd like to remind about
(1) the review for this,
(2) the access rights to AppVeyor,
(3) that since those few days, with no apparent change, this happens:

ben-farnham:
Just about to add you to appveyor...

ben-farnham:
OK, could you try logging in to appveyor now, you should have admin rights

ben-farnham:
As regards the root problem, not sure. Perhaps related to the python install on the build host
https://stackoverflow.com/questions/26828206/importerror-no-module-named-enum

The build yml file explicitly states
```python -m pip install enum34```
Is that the correct package?

Generally as regards appveyor/windows builds... I agree appveyor is far from perfect, I can't use the same docker approach as I used for the travis builds sadly - this means appveyor build hosts 'come as they are' (whereas travis build hosts are well specified).

I need quasar windows support: that's a core operational requirement. I didn't find anything better than appveyor.

pnikiel:
yes, that is the correct package...
I believe that nothing has changed that would prevent this from working. I will spend 15 minutes on this and If I fail, I will need your support in order to move on.

pnikiel:
Hi @ben-farnham, this has finally passed, thank you for your help!
Could I ask you for a review?

pnikiel:
Thanks!

PR #75, opened 2018-10-29
OPCUA-183 cmake3

pnikiel:
Tested on a simple server with open6 backend, did fine.


pnikiel:
OK both passing.
Ready to merge i suppose?

PR #76, opened 2018-10-29
Fixes for API changes in open62541 0.3

pnikiel:
A tiny part of open62541 API is used directly in quasar sources, this part adapts for the API changes between 0.2 and 0.3.
The code is ifdefed depending on open62541 defined version with the intent of moving entirely to 0.3 in foreseeable future.


pnikiel:
Hi,
I think today 7 days pass since the review was requested, and it does look pretty trivial.

pnikiel:
Haven't seen any interest in this review; merging.

PR #78, opened 2018-11-12
Opcua 1148 add build information to meta

ben-farnham:
Here we go folks - as per ticket OPCUA-1148 this adds
* build host
* built timestamp
* commit ID (git/svn)
* toolkitlibs

to both
* Meta.BuildInformation
* the '--version' command line output.

I haven't been able to test the SVN stuff from end to end (no SVN repos). I did try the python excerpt from version_control_interface.py :: get_latest_repo_commit for SVN and it appears to work using a stand-alone python prompt.


ben-farnham:
Ah, hang on, travis-CI failing. I'll find out why...

CMakeLists.txt
```diff
@@ -56,6 +56,12 @@ include_directories (
 	${PROJECT_BINARY_DIR}/Configuration
 	${SERVER_INCLUDE_DIRECTORIES}
 	)
+
```
pnikiel: Is this change necessary?

FrameworkInternals/meta_build_info.py
```diff
@@ -0,0 +1,55 @@
...
+	generatedFile.write('#define GEND_BUILD_TIME "{0}"\n'.format(now))
+
+def generateBuildHost(generatedFile):
+	build_host = socket.gethostname()
+	generatedFile.write('#define GEND_BUILD_HOST "{0}"\n'.format(build_host))
+
+def generateCommitID(generatedFile):
+	commitID = "No commit ID found"
+	try:
+		project_root_path = os.path.dirname(os.path.realpath(__file__))+os.path.sep+'..'
```
pnikiel: I'd suggest passing value of PROJECT_SOURCE_DIR from master CMakeLists into here (via command line arg or so).  I'm not sure if this assumption really scales well.

Meta/config/Meta.xsd
```diff
@@ -30,6 +30,7 @@
 			<xs:element name="Log" type="tns:Log" minOccurs="0" 	maxOccurs="1" />
 			<xs:element name="SourceVariableThreadPool" type="tns:SourceVariableThreadPool" minOccurs="0" 	maxOccurs="1" />
 			<xs:element name="Server" type="tns:Server" minOccurs="0" 	maxOccurs="1" />
+			<xs:element name="BuildInformation" type="tns:BuildInformation" minOccurs="0" 	maxOccurs="1" />
```
pnikiel: What motivates having BuildInfo in the config file schema?
To my mind it should not be there. Or?

Meta/include/meta.h
```diff
@@ -25,6 +25,18 @@
 #include <DRoot.h>
 #include <ASNodeManager.h>
 #include <DStandardMetaData.h>
+#include <ASNodeQueries.h>
 
 Device::DStandardMetaData* configureMeta(Configuration::Configuration & config, AddressSpace::ASNodeManager *nm, UaNodeId parentNodeId);
 void destroyMeta (AddressSpace::ASNodeManager *nm);
+
+template<typename TAddressSpaceType>
```
pnikiel: Pure cosmetics, If we could be consistent with already existing convention of template class names in quasar which doesnt use "T" as prefix, that would be better.

Meta/src/meta.cpp
```diff
@@ -158,6 +160,19 @@ void configureSourceVariableThreadPool(const Configuration::SourceVariableThread
     MetaUtils::linkHandlerObjectAndAddressSpaceNode(dSourceVariableThreadPool, asSourceVariableThreadPool);
 }
 
+void configureBuildInformation(const Configuration::BuildInformation& config, AddressSpace::ASNodeManager *nm,  AddressSpace::ASStandardMetaData* parent)
```
pnikiel: Same question here as few questions above - I don't understand why this purely runtime information is routed via Configuration objects.

Meta/include/meta.h
ben-farnham: done

CMakeLists.txt
ben-farnham: Yep - necessary. Needed to move the resolution of the OPCUA_TOOLKIT_LIBS variable (based on whatever is defined in the build config) before the inclusion of the Meta sub-directory. Without this Meta doesn't know what the toolkit libs are to publish via BuildInformation.

The block above the new location of OPCUA_TOOLKIT_LIBS  resolution (not moved in this branch) already references variable OPCUA_TOOLKIT_PATH - defined in the build config.

FrameworkInternals/meta_build_info.py
ben-farnham: Yep, I'll try that. The existing code is a little nasty.

Meta/config/Meta.xsd
ben-farnham: Yep, probably a good point. I'll try removing it (+references in the code to it). More soon...

Meta/src/meta.cpp
ben-farnham: and same answer (yep - it's a fair point)

Meta/config/Meta.xsd
ben-farnham: Done

Meta/src/meta.cpp
ben-farnham: Done

FrameworkInternals/meta_build_info.py
ben-farnham: Done
Meta/CMakeLists.txt
```
COMMAND python ${CMAKE_CURRENT_LIST_DIR}/../FrameworkInternals/meta_build_info.py --project_root_dir ${CMAKE_SOURCE_DIR} --target_generation_dir ${    PROJECT_BINARY_DIR}/generated --toolkit_libs "${OPCUA_TOOLKIT_LIBS}"
```

Then in meta_build_info.py
```
def process_args():
print 'meta_build_info.py, processing command line args...'
parser = argparse.ArgumentParser(description="""meta_build_info.py generates build info. Requires 3 args - Abs path to project root, abs path to th    e target generation directory (must already exist) and the toolkit libs""", formatter_class=RawTextHelpFormatter)
parser.add_argument('--project_root_dir', required=True, help='absolute path to quasar project root directory')
...

#used here

def generateCommitID(generatedFile, project_root_dir):
commitID = "No commit ID found"
try:
        commitID = version_control_interface.VersionControlInterface(project_root_dir).get_latest_repo_commit()
```

ben-farnham:
Ready to merge I think:
- feature works
- review points addressed (thanks Piotr - good points raised)
- CI builds green

Any objections?

ben-farnham:
Sure. No rush from my side, let me know how it goes...

pnikiel:
Hi @ben-farnham ,
first test with a 'regular' (though complex) quasar server (i.e. no yocto or so). Goes smoothly until it tries to compile meta.cpp:

/home/pnikiel/gitProjects/ScaOpcUa/Meta/src/meta.cpp: In function ‘void configureBuildInformation(AddressSpace::ASNodeManager*, AddressSpace::ASStandardMetaData*)’:
/home/pnikiel/gitProjects/ScaOpcUa/Meta/src/meta.cpp:170:133: error: ‘AS_TYPE_BUILDINFORMATION’ is not a member of ‘AddressSpace::ASInformationModel’
     AddressSpace::ASBuildInformation *asBuildInformation = new AddressSpace::ASBuildInformation(parent->nodeId(), nm->getTypeNodeId(AddressSpace::ASInformationModel::AS_TYPE_BUILDINFORMATION), nm,

looks like sth that should be easily reproducible at your end?

pnikiel:
aha! got a quasar bug, ASInformation generation is not dependent on the xslt (And it should be). Something to fix for me. Don't pay attention, I'm carrying on testing.

pnikiel:
OK 1st test results:
* build went OK (apart from unrelated quasar bug),
* in the address space everything is perfect,
* from the command line perfect as well.

will try on a petalinux project.

pnikiel:
Perhaps I could also ask you to include copyright header to meta_build_info.py ?

ben-farnham:
Done!

pnikiel:
Hi @ben-farnham,

I tested your contribution on PetaLinux(so Yocto). I realized I don't have a recipe for pygit (so it says that commit id is not found) for which I created a follow-up ticket to be resolved later.

Results: builds fine, runs fine. 

Thank you!

I believe this deserves merging.

ben-farnham:
Squash and merge do you think?

pnikiel:
Squash could be appreciated but I don't insist on it (i.e. the branch doesn't look particularly dirty to me).

ben-farnham:
Yep, I think a squash too. Thanks for testing.

PR #79, opened 2018-11-14
Implemented OPCUA-1211

pnikiel:
https://its.cern.ch/jira/browse/OPCUA-1211

### DIFF
--- a/Documentation/quasarCommands.html
+++ b/Documentation/quasarCommands.html
@@ -254,6 +254,25 @@
     .svnignore file, so the generated files will be ignored in your svn
     repository.<br>
     <br>
+    <h2><a name="symlink_runtime_deps"></a>symlink_runtime_deps</h2>
+    <p>New in quasar 1.3.0.</p>
+    <p>When you build out-of-source (which is the default since quasar
+      1.3.0) your executable will be done in build/bin (the default) but
+      some files (like ServerConfig.xml) which are normally in bin/ will
+      be needed to run it.<br>
+      This convenience function can be run:<br>
+      <br>
+    </p>
+    <ul>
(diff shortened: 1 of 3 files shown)


ben-farnham:
Or (thinking further) how about extending it a bit...

Call it _symlink_runtime_dependencies_ or something expressive and have it accept an open ended list of file paths. The script loops through the list and symlinks them to the (source-dir)\bin (and barfs if any of the file paths are not found).

A little late for this kind of suggestion I know. Better late than never (?)

pnikiel:
It does look like a good idea to me. What about the name you suggest and then:
* no arguments -> symlinks ServerConfig.xml and config*.xml
* one argument -> symlinks whatever wildcard-matches the argument?

ben-farnham:
I like the no-args idea.

Could we have multiple wildcard-matches?  To help developers with servers with other dependencies, maybe dependencies that are not say*.xml etc.

pnikiel:
Could you provide an example?
What I have already will match ANY PATTERN (including \* to match everything) in bin/ directory and symlink it to build/bin.

ben-farnham:
I just mean to symlink files that are maybe outside of SOURCE_DIR/bin to the BINARY_DIR/bin

ben-farnham:
symlink_runtime_dependencies(/way/out/west/somefile.ini, /way/out/east/someotherfile.txt)

pnikiel:
That seems pretty vague to me.
Since I don't really understand what I'd be implementing (and for whom, and what are the requirements etc) I'd prefer that your point is addressed by another ticket.
I've taken into account your points, also added documentation entry, will merge soon.
The command name got abbreviated: "symlink_runtime_deps" just to minimize the risk of us guitarists not being able to play from excessive keyboard strain.

PR #80, opened 2018-11-15
implementted

pnikiel:


pnikiel:
The context is UaO which seems the only case (so far) where we need to supply two additional arguments: class name and c++ namespace. But the change is of generic nature and we may sooner or later profit from it. Thanks for the review!

PR #84, opened 2018-12-19
Implemented shared object creation

pnikiel:
The motivation to this is to have an option to build a (complete) quasar-based server as a shared object. This is required to load it, run it (and do data exchange with it) from a higher-level programming language like Python or so.


pnikiel:
Ah, that is clever. Suggestions implemented. Merging.

PR #85, opened 2019-01-14
Now both backends use the same mechanism - g_RunningFlag

pnikiel:
1. Unified shutdown mechanism between both backends,
2. added ShutDown() function so all interfaces now have clean API to shut the server down without dealing with its internal data

### DIFF
--- a/Server/include/shutdown.h
+++ b/Server/include/shutdown.h
@@ -16,6 +16,8 @@
 #ifndef __SHUTDOWN_H__
 #define __SHUTDOWN_H__
 
+#include <uavariant.h>
+
 #ifdef _WIN32
 #  ifdef _WIN32_WCE
 /* Windows CE */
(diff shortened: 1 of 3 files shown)


Server/include/shutdown.h
```diff
@@ -16,6 +16,8 @@
 #ifndef __SHUTDOWN_H__
 #define __SHUTDOWN_H__
 
+#include <uavariant.h>
```
ben-farnham: Is this include needed? I didn't see any new references to variant types in the changes for this file.

Server/src/shutdown.cpp
```diff
@@ -28,11 +23,7 @@
 #  include <windows.h>
 # endif
 
-volatile unsigned int  g_ShutDown = 0;
-
-#ifdef BACKEND_OPEN62541
-UA_Boolean g_RunningFlag = 1;
-#endif
+OpcUa_Boolean g_RunningFlag = 1;
```
ben-farnham: Should the running flag be atomic? [This stackoverflow thread](https://stackoverflow.com/questions/16111663/do-i-have-to-use-atomicbool-for-exit-bool-variable?noredirect=1&lq=1) seems to suggest so.

Server/include/shutdown.h
ben-farnham: ok, fair enough

Server/src/shutdown.cpp
ben-farnham: Sounds fair enough

PR #90, opened 2019-02-14
symlink creation in new build/bin dir

parasxos:


parasxos:
Hi Piotr,
thanks. I put the relevant call in the suggested place but now it is never executed. The reason is that in the following code:

```
43	transformByKey(TransformKeys.AS_CMAKE, {'context':context})
	transformByKey(TransformKeys.D_CMAKE, {'context':context})
	...
48	if not os.path.isdir(projectBinaryDir):
49		print("PROJECT_BINARY_DIR {0} doesn't exist -- creating it.".format(projectBinaryDir))
50		os.mkdir(projectBinaryDir)
51	os.chdir(projectBinaryDir)
```

the program never enters line 49 and 50 as the dir is already made in line 43.

parasxos:
How is it now? I think it makes better sense

pnikiel:
Hi,

I tested and it works OK. I'd however like to ask for motivation of moving the seemingly unrelated code down? (where D_CMAKE, AS_CMAKE is generated etc)

pnikiel:
Apparently however there is an issue in windows:
Traceback (most recent call last):
  File "quasar.py", line 81, in <module>
    callee( makeContext(), *args )  # pack arguments after the last chunk of the command	
  File "C:\workspace\OPC-UA\quasar-test-project\FrameworkInternals\generateCmake.py", line 50, in generateCmake
    symlinkRuntimeDeps(context, '*.xml')
  File "C:\workspace\OPC-UA\quasar-test-project\FrameworkInternals\manage_files.py", line 524, in symlinkRuntimeDeps
    symlinkIfNotExists(config_file, os.path.join(context['projectBinaryDir'], 'bin', os.path.basename(config_file)))
  File "C:\workspace\OPC-UA\quasar-test-project\FrameworkInternals\manage_files.py", line 500, in symlinkIfNotExists
    os.symlink(src, dst)
AttributeError: 'module' object has no attribute 'symlink'


which means that merging this will kill any quasar build.
Ideas?

parasxos:
The motivation is that "transformByKey(TransformKeys.AS_CMAKE, {'context':context})" already creates the build dir in order to put AddressSpace/cmake_generated.cmake in place.

So if those lines are not moved down, then the function symlinkRuntimeDeps will never be executed.

pnikiel:
@parasxos thanks! I didn't know about it ;-)

So what remains is Windows ...

parasxos:
Hi @ben-farnham, could you please have a look on the discussion above on the windows "issue" with symlinks? 
It appears that symlinking needs elevated privileges in windows. Could you takeover the windows part?

parasxos:
Now that the windows issue is resolved, could you please close this PR?

PR #91, opened 2019-02-14
Implemented better handling when Variant is Null;

pnikiel:
This is relevant to all situations when a calc var is created on a hardware-polled value which can be NULL e.g. because of connectivity malfunction or so. In such a case WRN was too much for a log level.

### DIFF
--- a/CalculatedVariables/src/CalculatedVariablesChangeListener.cpp
+++ b/CalculatedVariables/src/CalculatedVariablesChangeListener.cpp
@@ -45,6 +45,13 @@ void CalculatedVariables::ChangeListener::operator ()(
     else if (newValue.value())
     {
         UaVariant variant (*newValue.value());
+        if (variant.isEmpty())
+        {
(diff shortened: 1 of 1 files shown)


meeludwig:
is it possible to distinguish between NULL value (=hw problems) and conversion to double problem (=design problem, bug) ? The message is clear, maybe another return code can be used ?

pnikiel:
Please If I can ask you not to close PRs.

pnikiel:
@meeludwig by design, CalculatedVariables can be only created on numeric variables (check CommonFunctions.xslt, fnc:isQuasarDataTypeNumeric).

In my understanding therefore, the message which I left with WRN covers the case of "design problem" (though in fact it would be like quasar design problem) whereas the one I added covers what you called "HW problem". 

Does it answer your question?

parasxos:
looks good

pnikiel:
Reviewed by Paris, and I see no further comments, merging.

PR #93, opened 2019-02-27
Implemented propagation of error code

pnikiel:


Server/src/opcserver_open62541.cpp
```diff
@@ -57,13 +57,20 @@ int OpcServer::start()
         throw_runtime_error_with_origin("UA_Server_new failed");
     m_nodemanager->linkServer(m_server);
     m_nodemanager->afterStartUp();
+    UA_StatusCode status = UA_Server_run_startup(m_server);
```
ben-farnham: Don't actually know what the call to ``` UA_Server_run_startup``` so this is a rater uninformed statement....

This code just reads rather weirdly: I mean we call a method ```m_nodemanager->afterStartUp();``` then immediately after call a method called ```UA_Server_run_startup(m_server)```. We are talking about different things starting here? Maybe it would help if we named variable ```status``` to give a further hint as to what just started.

Server/src/opcserver_open62541.cpp
```diff
@@ -74,7 +81,15 @@ int OpcServer::stop(OpcUa_Int32 secondsTillShutdown, const UaLocalizedText& shut
...
-    UA_Server_run(m_server, &g_RunningFlag);
+    while (g_RunningFlag)
+    {
+        UA_Server_run_iterate(m_server, true);
+    }
+    UA_StatusCode status = UA_Server_run_shutdown(m_server);
+    if (status != UA_STATUSCODE_GOOD)
+    {
+        LOG(Log::ERR) << "UA_Server_run_shuttdown returned not-good. Error was:" << UaStatus(status).toString().toUtf8();
+    }
```
ben-farnham: Might be nice to also log a message if the status is good. Shutting down the server sounds like a key moment in the server lifetime; sometimes it's nice to read a log message that (at this point anyway) the server was still on the happy path.

Server/src/opcserver_open62541.cpp
pnikiel: Yes, the node manager is obviously a different thing than the server, so is there any confusion here really?

Server/src/opcserver_open62541.cpp
pnikiel: OK, I don't mind, can be done.

pnikiel: Pushed.

pnikiel:
Are there any subsequent questions/comments or can this be merged?

ben-farnham:
all good from my side

PR #95, opened 2019-03-23
Implemented setting arrayDimensions for arrays

pnikiel:
ArrayDimensions attribute will be set.
In future we can think how to also propagate array dimensions restrictions from Design.

### DIFF
--- a/AddressSpace/designToClassBody.xslt
+++ b/AddressSpace/designToClassBody.xslt
@@ -221,6 +221,9 @@ UaVariant v;
 			UaVariant v;
 			<xsl:value-of select="fnc:convertVectorToUaVariant('vect', 'v', @dataType)"/>
 			m_<xsl:value-of select="@name"/>-&gt;setValueRank( 1 );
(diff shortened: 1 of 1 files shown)


meeludwig:
this makes arrays work again for open6. Thanks !
I'm fixing my project now, if sth related to arrays pops up again we'll make a separate branch/ticket.
Thanks also for creating https://its.cern.ch/jira/browse/OPCUA-1380 and solving it, I generate these extra and superfluous config loaders somewhere. I understand that a few switches in the xsd should suppress them for the "initializeWith=ValueAndStatus cases.
ready to merge a.s.a.p.

**array dimension protection at design:**
array dimensions are protected at runtime with min/max, as stated when we merged the arrays into quasar. Since arrays are dynamical in size during runtime, the only thing the design protection would actually bring as a new benefit would be the better restricted Configuration.xsd. This would prevent  the user from validating a config.xml (for population) which contains an array-out-of-bounds. Otherwise it would not change anything.
So I would say array-dim-protection by design is "nice to have" but low prio.

PR #97, opened 2019-03-26
Implemented default values of config entries

pnikiel:
Taken care of the following aspects:
* Design schema,
* propagation to Config schema,
* Design Validation (2nd stage validator),
* Config documentation

Everything should be backwards-compatible.


pnikiel:
Thank you Gentlemen.

PR #99, opened 2019-04-03
implemented copyIfNotExists to handle windows implementation of symli…

ben-farnham:
…nk_runtime_deps command

### DIFF
--- a/FrameworkInternals/manage_files.py
+++ b/FrameworkInternals/manage_files.py
@@ -24,6 +24,7 @@ import errno
 try:
 	import sys
 	import os
+	import platform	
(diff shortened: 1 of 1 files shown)


pnikiel:
Added @parasxos

pnikiel:
Hi,
one very general comment: it deserves to put a comment which would explain why in windows we call copy instead of symlink (elevated privileges required + no .symlink in python 2.7 stdlib).

pnikiel:
Hi @ben-farnham ,
I gave it a spin on our rather complex OPC-UA server, to my mind it is OK.

PR #102, opened 2019-06-05
fixed upgrade_design which didn't work after transforms rework

pnikiel:


pnikiel:
Hi @ben-farnham ,
context contains a number of build-related settings (e.g. source dir, bin dir, etc) and other things.
It is created in quasarCommands.py (assuming a command is called from quasar.py) and it is consumed in multiple places, e.g. see transformByKey()

thanks for the review

PR #107, opened 2019-08-28
OPCUA-1472 support current object in calculated variables, generic formulas and constants

pnikiel:


pnikiel:
FYI: travis build fails because, as I learnt, BOOST_LIBS is configured system-wide, dunno why - will talk about it with Ben.
When tried manually from Docker it builds fine.

pnikiel:
Hello Paris,

thank you for your effort on this. Since boost::regex is used in multiple places in quasar,
I registered another ticket:

https://its.cern.ch/jira/browse/OPCUA-1492

My preference would be to address migration from boost::regex to std::regex holistically in quasar; I'd plan this to happen also for quasar-1.3.5 so build configuration files would not need to be changed.

As for the constants, once I finish OPCUA-1500, I'll withdraw from the concept.

pnikiel:
Dear Paris,

after some considerable effort, I realized that std::regex is useless in gcc4.8.x - the compiler is simply stone-age. More discussion will follow in:
https://its.cern.ch/jira/browse/OPCUA-1492

PR #108, opened 2019-08-28
Made m_addressSpaceLink private

pnikiel:
This was planned for ages...
Device classes should use getAddressSpaceLink() instead of m_addressSpaceLink,
this is especially important for code that gets executed at init or at destruction (especially if that code was spawned from an extra thread) because the risk of nullptr dereference is minimized.


pnikiel:
Thanks!

PR #109, opened 2019-08-30
A working version but not so clear and clearly skipping device instan…

pnikiel:
…tiation, etc


pnikiel:
Thanks @parasxos !
@ben-farnham , please open a ticket with follow-up requests if needed (the basic philosophy is here already).

PR #110, opened 2019-09-02
Opcua 1472 support current object in calculated variables via xpressive

pnikiel:


pnikiel:
Constants were removed, and the shift to boost::xpressive was done not to require additional libs. Gcc 4.8 std::regex was found to be useless.

PR #111, opened 2019-09-02
OPCUA-1434: escape double quotes in the commitID string returned from pygit2.Repo…

ben-farnham:
…sitory.describe() call


ben-farnham:
Not the biggest change, it's true. Merging...

PR #114, opened 2019-09-19
Opcua 1497 varied level parents

pnikiel:
So now we review the support for $thisObjectAddress (i.e. same as $_ ) and $parentObjectAddress,
both were already successfully tested.

### DIFF
--- a/CalculatedVariables/src/CalculatedVariablesEngine.cpp
+++ b/CalculatedVariables/src/CalculatedVariablesEngine.cpp
@@ -99,8 +99,15 @@ static std::string elaborateParent(
     {
         cutOffIndex = input.rfind('.', cutOffIndex);
         if (cutOffIndex == std::string::npos)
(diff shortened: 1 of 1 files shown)


CalculatedVariables/src/CalculatedVariablesEngine.cpp
```diff
@@ -87,6 +87,33 @@ double* CalculatedVariables::Engine::parserVariableRequestHandler(const char* na
...
+    {
+        cutOffIndex = input.rfind('.', cutOffIndex);
+        if (cutOffIndex == std::string::npos)
+            LOG_AND_THROW_ERROR(thisFormulaAddress, "Not enough levels to go up!"); 
+        levels--;
+        if (levels>0)
+        {
+            if (cutOffIndex>0)
+                cutOffIndex--;
+            else
```
parasxos: It seems that lines 101-102 already cover you from this else condition to be happening.

PR #116, opened 2019-09-20
Implemented missing header

pnikiel:
It was found out that for long time we had that issue that the body of the member function was generated, but it was not in the headers... now it is fixed!


pnikiel:
thx Ben!

PR #118, opened 2019-10-21
Ninja builds support

pnikiel:


pnikiel:
Thank you @parasxos .
I implemented your remarks. 
Please note that in order to run Ninja now you will use
--builder Ninja (note capital N)
as it is directly passed to CMake.
(i.e. it is now agnostic of the builder - this all gets passed to CMake directly)

PR #120, opened 2019-10-24
Implemented OPCUA-739 (Meta amalgamation)

pnikiel:
Pasting the description from what I wrote in Meta's CMakeLists:

Meta, as its 'first approach' is simply made by quasar itself, i.e. a quasar server was once made 
with a Design corresponding to what one sees as StandardMetaData. The address-space and device sources,
and XSD schema was exported from the project, and along certain custom-made additions made it to the Meta module.
This structure is not very build-time friendly though; a total of 20+ source files are compiled where 99% is
composed of same included headers and less than 1% is the real code. Therefore it pays off to amalgamate files
sharing same headers; with an 'empty' quasar server this yields a build time reduction of about 2x (significant!).
The reason that MetaBuildInfo.cpp is not amalgamated is that it depends on a file which is regenerated at each build
and there is no reason to rebuild amalgamated files without the factual reason!


FrameworkInternals/quasar_busybox.py
```diff
@@ -0,0 +1,45 @@
+'''
+quasar_busybox.py
```
ben-farnham: Why call this file 'busybox.py' ? I would have thought this python function is a candidate for addition to quasar_basic_utils.py or maybe manage_files.py

pnikiel: The inspiration for the name came from the BusyBox:
https://en.wikipedia.org/wiki/BusyBox
I believe this is a common name in the software engineering world, but we can discuss it.

Why not quasar_basic_utils or manage_files? I thought that neither of them actually delivers(-ed) functions which could be independent functions callable directly from the shell (thus, available directly to cmake). I preferred to keep separation between what is internal to quasar's python-based scripting (i.e. typically called via quasar.py) and what is external; the latter was not represented yet so that's why I created a new file. 

So keeping in mind the separation, I think that adding a new file was justified.

Let me know your viewpoint.

ben-farnham: busybox sounds unix/linux-y to me, so, unlikely to be a place you'd look into if you're a windows developer poking about in quasar. Windows == first class quasar citizen and all that.

ben-farnham: Propose move to manage_files.py

pnikiel: as per Windows comment... and 1st/2nd class citizen... well.. not sure about it, to be checked.

But for the organization reasons it remains a valid question. @parasxos , what is your viewpoint on it, since you already reviewed this part?

parasxos: To my mind a new file is justified if it has a clear separate scope different than the existing ones.

The name of the library could be descriptive of its scope, as it is the developers that are going to use it and our lives to be made easier. 

Honestly, I didn't think of Ben's concern when reviewing. I think that a more descriptive OS-agnostic name could be used to ease that concern e.g. why not consider manage_files_external.py or utils_external.py or something similar implying the different scope?

Would that be enough?

pnikiel:
OK, the file was renamed from quasar_busybox to quasar_utils.
Thank you both.

PR #125, opened 2019-11-18
Run all py files with autopep8 --select E101,W291,W293

pnikiel:
FYI:
if you look at this file in this way:
https://github.com/quasar-team/quasar/pull/125/files?utf8=%E2%9C%93&diff=unified&w=1
It will tell you those are only whitespace changes.


pnikiel:
@ben-farnham , while I consider your comment quite humorous, I'm not entirely sure if you realized that we had a mix of indentations between different lines of the same .py file. 
That is considered an error in Python3. This fixes one of the errors preventing quasar being used in Python 3.

ben-farnham:
I realized, I still think your whitespace is better.

pnikiel:
haha ;-)

PR #127, opened 2019-11-25
Opcua 1610 python3 appveyor build

ben-farnham:
What do you think folks? Fixes the windows build inline with python3. Only contentious issue might be me replacing `variable_value is test_value` which I understand is actually an _identity_ (same object) test with `variable_value == test_value` which is an equality test.

I like mine better. But then I would say that.


pnikiel:
Thanks, I agree it's a good change.

PR #129, opened 2019-12-02
quasar-side of OPCUA-1642 implemented

pnikiel:
Hi,
this is generally related to OPCUA-1644 from open62541-compat.
It follows the move of UA_Server business from quasar to open62541-compat, so quasar becomes independent of open62541 internals related to UA_Server.
OPCUA-1644 was already reviewed and was released as open62541-compat 1.1.5


pnikiel:
Hi @ben-farnham ,
generally nothing is lost as all of the previous logging is now in the respective methods of open62541-compat, so the net effect is the same (and to my mind, there it is enough). I hope this sounds satisfactory.

PR #136, opened 2019-12-18
simplified cert, provide example code for handling .ua cert store

meeludwig:
as discussed yesterday:
- simplified cert (a certificate (der) & private key (pem) ) behavior
- provide example code for handling .ua cert store, 
- handle only first endpoint cert,
- keep calls to class Certificate disabled in meta.cpp
- keep dragging dead code as example, to help server developer to implement at own initiative
- only relevant to .ua toolkit, open6 does not handle certs anyway for now in quasar
- builds for .ua and .open6, no further tests

### DIFF
--- a/Meta/include/Certificate.h
+++ b/Meta/include/Certificate.h
@@ -26,21 +26,12 @@
 /// load
 class Certificate {
 public:
-    /// behavior:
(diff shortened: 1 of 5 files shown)


meeludwig:
rebase to branch OPCUA-1541-certificates2, forget this one

PR #142, opened 2020-02-17
Capture group now finishes with encountering a parenthesis... (extended \S catch)

pnikiel:
Description of the problem being solved:
https://its.cern.ch/jira/browse/OPCUA-1726


pnikiel:
Thanks for the suggestion. I believe there are many incremental improvements like that to be done, but in my viewpoint they'd be better dealt with when we can settle on a compiler that does support C++11 fully, which would permit normal std::regex instead. So I'd merge it as is for now.

PR #146, opened 2020-03-09
Opcua 1756 build system improvements

schlenk4:


CMakeLists.txt
```diff
@@ -39,6 +39,8 @@ endif()
 
 include( ProjectSettings.cmake )
 
+include( FrameworkInternals/BoostSetup.cmake )
```
ben-farnham: As discussed in today's meeting, to handle situations where `find_package(Boost ...)` does not work so well, I would like to insert my own boost locator file. So I propose
```
if( NOT DEFINED BOOST_LOCATOR_CMAKE_FILE)
  set( BOOST_LOCATOR_CMAKE_FILE FrameworkInternals/BoostSetup.cmake )
endif()
message( STATUS "locating boost using cmake file [${BOOST_LOCATOR_CMAKE_FILE}]" )
include( $BOOST_LOCATOR_CMAKE_FILE )

Comment (ben-farnham) [reply CMakeLists.txt]: So in my build configs I would 
```
set( BOOST_LOCATOR_CMAKE_FILE FindBoostWithBen.cmake )
```
And in those files I must
1. add boost include directories with `include_directories`
2. set variable `BOOST_LIBS`

Comment (schlenk4) [reply CMakeLists.txt]: Ok, thanks for the proposal. I will change the var name slightly since it is not only about the location but also the libs needed by quasar.

**Comment by ben-farnham** [comment]
Just tried to update the CAEN server using this branch. Got this error
image

**Comment by ben-farnham** [comment]
Doh! Ignore that. Forgot to clone recursively I guess

File: CMakeLists.txt [hunk 2]
```diff
@@ -32,59 +32,39 @@ add_definitions(
...
-	  include( ${PROJECT_SOURCE_DIR}/FrameworkInternals/build_config_selector.cmake )
+    include( ${PROJECT_SOURCE_DIR}/FrameworkInternals/build_config_selector.cmake )
 else()
-     message(FATAL_ERROR "Build configuration has not been chosen yet. Please run 'quasar.py set_build_config'")
+    message(FATAL_ERROR "Build configuration has not been chosen yet. Please run 'quasar.py set_build_config'")
 endif()
 
 include( ProjectSettings.cmake )
 
+if( ( NOT DEFINED IGNORE_BOOST_SETUP ) OR ( IGNORE_BOOST_SETUP STREQUAL "OFF" ) )
```
ben-farnham: Please can variable `IGNORE_BOOST_SETUP` be changed to `IGNORE_DEFAULT_BOOST_SETUP` or something like that. Otherwise the variable is rather misleading, since no matter what to build a quasar server there must be _some_ boost setup. The question is rather whether the developer opts for the standard boost resolution, or uses a handmade one.

ben-farnham: otherwise looks good to me.

ben-farnham: CAEN server builds on windows too with these changes (needs one line change to build configs to instruct main CMakeLists.txt to ignore default boost setup)

pnikiel:
This one was merged in three separate chunks into quasar 1.3.1. Closing.

PR #147, opened 2020-03-11
OPCUA-1541: reading cert paths from server config if .ua toolkit is u…

meeludwig:
**OPCUA-1541**

- reading cert paths from server config if .ua toolkit is used
- cleanup of Certificate class
- example call as comment in meta.cpp:202, otherwise the certificate class is unused by quasar default


meeludwig:
sorry, left a debugging line dangling. check CI again

PR #148, opened 2020-03-11
Opcua 1541 certificates2

meeludwig:
**OPCUA-1541**
- reading cert paths from server config if .ua toolkit is used
-     cleanup of Certificate class
-     example call as comment in meta.cpp:202, otherwise the certificate class is unused by quasar default
- left debug code dangling, fixed
CI is fine (open6 only as it seems)

### DIFF
--- a/Meta/include/Certificate.h
+++ b/Meta/include/Certificate.h
@@ -26,21 +26,12 @@
 /// load
 class Certificate {
 public:
-    /// behavior:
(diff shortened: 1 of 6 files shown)


pnikiel:
Can it be closed until it works?

parasxos:
Closing this as per Michael's request in OPCUA-1541.

PR #151, opened 2020-03-16
Rename open6_win_configuration.cmake to open62541_win_config.cmake

ben-farnham:
branch for windows CI build fix for feature branch _OPCUA-1756_build_system_improvements_


ben-farnham:
OK, done - appveyor & travis build happy. Ready for review & merge to your feature branch if you agree with the changes I made here.

Was slightly more involved than expected (see main CMakeLists.txt file) because I actually handle boost inline in the windows CI build configuration file. Which I think is an OK way of doing things. I made the changes to CMakeLists.txt to handle that.

CMakeLists.txt
```diff
@@ -42,8 +42,15 @@ include( ProjectSettings.cmake )
 if( NOT DEFINED BOOST_SETUP_CMAKE_FILE )
   set( BOOST_SETUP_CMAKE_FILE FrameworkInternals/BoostSetup.cmake )
 endif()
-message( STATUS "Setting up boost location and libraries using [${BOOST_SETUP_CMAKE_FILE}]")
-include( ${BOOST_SETUP_CMAKE_FILE} )
+if( EXISTS ${CMAKE_CURRENT_LIST_DIR}/${BOOST_SETUP_CMAKE_FILE} )
```
schlenk4: Is this really necessary? Looks like a redundant fallback. The first one should be sufficient, or?

ben-farnham: in the windows CI build config file I do this
```
set( BOOST_SETUP_CMAKE_FILE HANDLED_INLINE_IN_BUILD_CONFIG )
```
So, I wanted the option that i could also tell the main CMakeLists.txt that there is no boost setup file to collect (boost is just resolved inline in that build config file)

schlenk4: Then it is probably easier to just check a variable if the boost setup should be ignored or not and if it is ignored hint that the user must be dealing with this in the build config.

PR #164, opened 2020-04-06
Implemented OPCUA-1764

pnikiel:


parasxos:
Hi Piotr, I do like it and it also seems ok. I trust that you tested it.
One point to consider in general for placeholder/hollow files is that while it is very welcoming for the developers that some documentation on how to fill them is there, I wouldn't claim responsibility for them (license/sign etc.) ;)
Thats my point of view. (also applies to the Device generated cpp files)

pnikiel:
Hi @parasxos,
while I mentally agree with you, the act of doing so was requested from us by the CERN Knowledge Transfer group; apparently just letting files go without such headers and copyrights is not 100% legal according to some rules and regulations.
So not much I could do there, I think.

parasxos:
Since we are mentally... paired :) one can find other ways to achieve the same i.e avoid using the reserved filename and provide the guidance/documentation in a file with a postfix (e.g _DeviceCustom_Example.cmake_ ) or/and document the option to use a custom cmake file for devices in the build documentation of quasar.

pnikiel:
@parasxos I'm totally fine with this. I was just a bit afraid that if this gets into .Example or documentation that the users won't really see it. What is your opinion @ben-farnham  on @parasxos comment?

parasxos:
Up to now, developers (including me) were looking to add those files in DeviceLogic CMakeLists.txt. 
One can put a pointer to the example there, such that it is equally easy to find as before ;)

pnikiel:
To me this is all fine. But just for my understanding: the problem we're solving here is that header+copyright info, right?

pnikiel:
@parasxos  your comment was implemented. Thank you!

PR #171, opened 2020-05-08
Added initialValue validation

pnikiel:


pnikiel:
a kind reminder

FrameworkInternals/DesignValidator.py
```diff
@@ -75,10 +76,25 @@ def count_children(parent, element_name):
...
-    except Exception:
+    except AttributeError:
         pass  # that's OK, it's simply not there
     return count
 
+def assert_numeric_literal_valid(literal, quasar_data_type, locator):
+    """Will throw if given literal is out of numeric limits"""
+    try:
+        if quasar_data_type in ['OpcUa_Float', 'OpcUa_Double']:
+            test_f = float(literal)
```
ben-farnham, line 87: Is it safe to transform double to float in python?

pnikiel: from https://docs.python.org/3/library/stdtypes.html

> There are three distinct numeric types: integers, floating point numbers, and complex numbers. In addition, Booleans are a subtype of integers. **Integers have unlimited precision.** **Floating point numbers are usually implemented using double in C**; information about the precision and internal representation of floating point numbers for the machine on which your program is running is available in sys.float_info. Complex numbers have a real and imaginary part, which are each a floating point number. To extract these parts from a complex number z, use z.real and z.imag. (The standard library includes the additional numeric types fractions.Fraction, for rationals, and decimal.Decimal, for floating-point numbers with user-definable precision.)

hope that is justified.

pnikiel:
Thanks @parasxos  and @ben-farnham

PR #172, opened 2020-05-08
Now its atomic, concurrent build wont break it

pnikiel:


pnikiel:
This one is relatively important as it sometimes breaks builds in concurrent build jobs envitonment.

pnikiel:
one week passes soon, I will merge it basing on my own taste soon if there is no review done.

PR #173, opened 2020-05-11
OPCUA-1798 catch cv array with initialValue

pnikiel:


parasxos:
It went in as part of OPCUA-1880. That PR is thus not valid.

PR #176, opened 2020-05-15
Moved exception handler from quasar.py to quasar_basic_utils to let o…

pnikiel:
…ther modules profit from it


pnikiel:
Thx @parasxos

PR #180, opened 2020-05-19
Opcua 1854 configuration decoration with content order

ben-farnham:
Here we go...


Configuration/Configurator.h
```diff
@@ -37,6 +40,31 @@ bool configure (std::string fileName,
         AddressSpace::ASNodeManager *nm, ConfigXmlDecoratorFunction
         configXmlDecoratorFunction = ConfigXmlDecoratorFunction()); // 'empty' function by default.
 
+template<typename TParent, typename TChildren, typename TChildTypeId>
+void validateContentOrder(const TParent& parent, const TChildren& children, const TChildTypeId childTypeId)
```
pnikiel, line 44: Deserves to remark:
1. that this check is only affecting configuration that got decorated and is expected to be transparent otherwise.
2. how does this algorithm actually work? could you add 2 or 3 phrases with an example of when it would throw?

Configuration/Configurator.h
```diff
@@ -25,7 +25,10 @@
 #define CONFIGURATOR_H_
 
 #include <string>
+#include <set>
+#include <numeric>
 #include <functional>
+#include <boost/format.hpp>
```
pnikiel, line 31: I'm not in favour of that, especially that there was so much effort to cut down boost to the minimum.
The requested functionality seems entirely possible in standard C++, perhaps a bit less hip but without growing dependencies.

Configuration/Configurator.h
```diff
@@ -37,6 +40,31 @@ bool configure (std::string fileName,
...
+      }
+      validChildIndices.erase(pos);
+    }
+  }
+  if(!validChildIndices.empty())
+  {
+    throw std::range_error(boost::str(boost::format("%1% ERROR parent has [%2%] child objects unregistered in content order") % __FUNCTION__ % validChildIndices.size()));
+  }
+};
+
```
pnikiel, line 67: Not fully reviewed yet, I'll redo after requested comments are added.

Configuration/templates/designToConfigurator.jinja
```diff
@@ -1,4 +1,4 @@
-{# © Copyright CERN, 2015.                                                      #}
+{# �� Copyright CERN, 2015.                                                      #}
```
pnikiel, line 1: Please fix your UTF-8 settings.

Configuration/templates/designToConfigurator.jinja
```diff
@@ -23,9 +23,6 @@
 {% set this = designInspector.objectify_root() %}
 {{ headers.cppFullGeneratedHeader() }}
 
-#include <xercesc/dom/DOMNodeList.hpp>
-#include <xercesc/dom/DOMElement.hpp>
```
pnikiel, line 27: Great sign!

Configuration/templates/designToConfigurator.jinja
```diff
@@ -98,85 +96,68 @@ using namespace CalculatedVariables;
...
-          configure{{innerClass}} (*innerObjectConfig, nm, {{parentNodeId}});
-      {% endif %}
-      continue;
+        break;
+      {% endfor %}
+      case Configuration::{{xsdParentType}}::CalculatedVariable_id:
+      {
+        const auto& xmlObj(config.CalculatedVariable()[xmlIndex]);
+        LOG(Log::DBG)<<__FUNCTION__<<" Configuring type [id:"<<xmlType<<", nm:CalculatedVariable] ordering index ["<<xmlIndex<<"]";
+        Engine::instantiateCalculatedVariable(nm, {{parentNodeId}}, xmlObj);
```
pnikiel, line 151: Could we NOT cut-off the namespace here? Prependign with CalculatedVariables:: will, IMO, but more readable.

FrameworkInternals/original_files.txt
```diff
@@ -108,6 +108,9 @@ File designToConfigValidator.jinja          must_exist,must_be_versioned, md5=ch
 File designToConfigurationXSD.jinja         must_exist,must_be_versioned, md5=check, install=overwrite
 File designToConfigurator.jinja             must_exist,must_be_versioned, md5=check, install=overwrite
 
+Directory Configuration/include install=create
```
pnikiel, line 111: FYI Configuration module is slightly special in that it includes *directly* from Configuration rather than from Configuration/include. If you insist on putting that file in Configuration/include then you should fix the master CMakeLists.

Configuration/include/ConfigurationDecorationUtils.h
```diff
@@ -0,0 +1,59 @@
...
+    * 2. Content Ordering: maintaining the ordering mechanism quasar uses to process
+    *    configuration elements in the correct order.
+    *
+    * @param parent: Parent tree node.
+    * @param children: The collection of children (owned by the parent) to which the
+    *        new child will be appended.
+    * @param child: The new child object
+    * @param childTypeId: The ordering type ID (as gen'd by xsdcxx) of the new child 
+    *        object          
+    */
```
pnikiel, line 47: This is great. It's so rare to see proper DoxyGen in this project ;-)

Configuration/Configurator.h
ben-farnham: Every child object (referenced by its index in the child object collections attached to the parent) should have a corresponding entry in the parent's content_order collection. This is the information xsdcxx's content_order mechanism runs on.

Objects arising from decoration or from the original XML parse are treated equally: I mean the algorithm passes over both new (decoration) objects and original (parsed) objects. This is transparent for original objects because it validates as above: Original objects conform to this algorithm by definition of the content_order mechanism.

As regards when/how might it fail: The documentation contains an example of how decoration can fail. Documentation/UserDefinedRuntimeConfigurationModification.html

Configuration/Configurator.h
ben-farnham: OK. It _is_ less hip though.

Configuration/templates/designToConfigurator.jinja
ben-farnham: Not sure I understand this point. Which line do you refer to ?

FrameworkInternals/original_files.txt
ben-farnham: I put the new file in Configuration/include as it's user facing: Configuration decoration code must include that file and I felt it better reinforced the quasar norm of putting include files in _include_ dir.

What do you think? Have a strong preference ?

Configuration/Configurator.h
ben-farnham: https://github.com/quasar-team/quasar/commit/7b4063ca76b244aa3c8c7906d292619c038108a3

Configuration/templates/designToConfigurator.jinja
ben-farnham: Done

https://github.com/quasar-team/quasar/commit/be9fc81d3ec56c2b4505f7dc5f07ab1540f4d616

ben-farnham:
Hi @pnikiel ,

Thanks for the review. Left a few responses/questions for you above and resolved everything else that was clear to me. I'll wait for your responses to continue.

Configuration/templates/designToConfigurator.jinja
pnikiel: line 168.

FrameworkInternals/original_files.txt
pnikiel: A-ha I checked and Configuration/include is already included by the master CMakeLists even though it did not exist so far.
OK, I agree, it could be added there.

Configuration/templates/designToConfigurator.jinja
ben-farnham: Ah I see (_using namespace_ was a straight copy from XSLT). Done
https://github.com/quasar-team/quasar/commit/cf549cbae2cbf1208186d8edaa979b265b6ddf94

pnikiel:
Hi @ben-farnham ,
I think it's quite good. I think it could enter the game (i.e. merge it).
I saw there seems to be some support for https://its.cern.ch/jira/browse/OPCUA-1873 , shall I test it against that one too?

ben-farnham:
Thanks for the review @pnikiel . 

Indeed OPCUA-1873 (GenericSNMP) should also be fixed in this branch: in the end it was too complicated to untangle the fix into a dedicated branch. If you could test GenericSNMP too that would be great. I tested generation (from design.xml) - i.e. device classes generated, with 'empty' logic - and compilation. That was OK, but if you could test with the complete server that would be a more complete test.

I'll merge after you report back as regards the GenericSNMP server.

pnikiel:
OK, also with GenericSNMP seems fine. Probably good to go.

PR #186, opened 2020-05-27
Added mutex validation

pnikiel:


FrameworkInternals/DesignInspector.py
```diff
@@ -110,24 +110,25 @@ def class_has_legit_device_parent(self, class_name):
...
-                  + Style.RESET_ALL)
-        return parent_structure
+            return None  # child class w/o DeviceLogic, so no device lo
+
+    def get_parent_struct(self, class_name):
+        """ Returns whatever should be typedef'ed as Parent_DX for X = class_name """
+        # a case where device logic parent cant be established
+        parent = self.get_parent(class_name)
+        if parent is None:
+            return 'struct{/*No exact Parent of the class*/}'
```
ben-farnham, line 129: This `struct {/*blah*/}` seems like C++ specific markup creeping in here. I'd thought DesignInspector was rather a language agnostic source of information, and transformation of DesignInspector information into the specifics of the final format (I guess C++ code here) is the job of the jinja templates.

What do you think ?

pnikiel: yes, I agree with you. Thanks for noticing. Since this 'creeping-in-of-C++' was already released (this is just a rework) I'd create an improvement ticket to address in one of upcoming releases. Thank you.

ben-farnham: OK - sounds good. No other comments: Looks good

PR #189, opened 2020-06-23
OPCUA-1881 removed parent struct generation from design inspector

pnikiel:


pnikiel:
Suggested in some earlier review comments, see OPCUA-1881 for more info

ben-farnham:
Yep, cleaner I think.

pnikiel:
I do concur, thanks.

PR #190, opened 2020-08-04
Removed UA_DISABLE_COPY from AS Headers and from ASNodeManager

pnikiel:
Mostly cosmetic improvements noticed on Jinja conversion.


pnikiel:
> Deleting the copy constructor and the copy operator will suppress the implicit declaration of a move constructor and move assignment operator. One could be explicit here for readability.
> Well known as "the rule of 5".

Good observation. I think it is correct - we do not want move ctr, move operator in device logic classes. Thank you guys.

pnikiel:
Note suggested by @parasxos added. thx.

PR #191, opened 2020-08-12
Opcua 1928 rework sourcevars jinja

pnikiel:


AddressSpace/templates/designToSourceVariablesBody.jinja
```diff
@@ -0,0 +1,411 @@
...
+        // Obtain Device Logic object
+        const AS{{className}}* addressSpaceObject (nullptr);
+        addressSpaceObject = dynamic_cast<const AS{{className}}*> ( m_parentObjectNode );
+        if (addressSpaceObject == m_parentObjectNode)
+        { /* OK. Proper cast. */
+          Device::D{{className}}* device = addressSpaceObject->getDeviceLink();
+          if (device != 0)
+          {
+            /* OK - Device Logic linked. */
+            {% if sv.get('addressSpaceReadUseMutex') == 'of_this_operation' %}
```
ben-farnham, line 114: Should this be made into a macro - looks like it repeats a few times in this source file

AddressSpace/templates/designToSourceVariablesBody.jinja
```diff
@@ -0,0 +1,411 @@
...
+  OpcUa_UInt32 hTransaction,
+  OpcUa_UInt32        callbackHandle,
+  const UaNode *parentNode
+)
+{
+  if (! sourceVariableThreads)
+    throw std::runtime_error("Attempted Source Variable operation, but Source Variable threads are not up.");
+  switch (jobId)
+  {
+    {% for className in designInspector.get_names_of_all_classes() %}
```
ben-farnham, line 301: another macro possible here? This logic (with all the jinja ifs and fors) looks pretty similar to the logic in the method below.

PR #192, opened 2020-08-14
OPCUA-1760 Removal of unnecessary headers

pnikiel:


ben-farnham (approved):
Looks good, always nice to see cruft go away

PR #194, opened 2020-08-21
Checking for astyle return value

pnikiel:


FrameworkInternals/transformDesign.py
```diff
@@ -185,7 +185,10 @@ def transformDesign(xsltTransformation, outputFile, requiresMerge, astyleRun, ad
 
         if astyleRun:
             try:
-                subprocess.call([getCommand('astyle'), outputFile])
+                return_code = subprocess.call([getCommand('astyle'), outputFile])
```
ben-farnham, line 188: any profit in having the return_code in the logged error message ?

ben-farnham (approved):

FrameworkInternals/transformDesign.py
pnikiel: I don't think so, astyle do not seem to have specific return codes.
But I can add it to the message, no big deal.
I understand there are no more remarks on this one.

PR #202, opened 2020-09-23
Logical condition handled in the other branch

pnikiel:
The problem (fixed here) prevents certain types of classes to be visualised correctly.


ben-farnham (approved):
looks good.

PR #211, opened 2020-10-01
Added server fixture

pnikiel:


parasxos (approved):
I give my blessing to this contribution despite line 73 and 75 in server_fixture.py contradict each other ;)

PR #232, opened 2020-11-12
OPCUA-2048 log server startup info (nebula)

pnikiel:


Server/src/BaseQuasarServer.cpp
```diff
@@ -49,10 +49,16 @@
...
-#ifndef __GNUC__
+#ifdef __linux__
+#include <unistd.h>
+#include <pwd.h>
+#elif _WIN32
 #include <windows.h>
+#include <Lmcons.h>
 #endif
 #include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/filesystem.hpp>
```
pnikiel, line 60: I'd have strong preference not to break existing expectations.

Server/include/BaseQuasarServer.h
```diff
@@ -99,10 +99,17 @@ class BaseQuasarServer
...
+    // Gets the current process environment variables
+    std::string getProcessEnvironmentVariables() const;
+    // Gets the current process owner
+    std::string getProcessOwner() const;
     //Cleans the server before exiting.
     void shutdownEnvironment();
     //Handler for initializing the node manager configuration only when the server is ready
     UaStatus configurationInitializerHandler(const std::string& configFileName, AddressSpace::ASNodeManager *nm);
 
+    std::list<std::string> m_commandLineArgs;
```
pnikiel, line 113: What is the reason for storing this information? Can't it be simply printed and forgotten?

Server/src/BaseQuasarServer.cpp
```diff
@@ -198,6 +204,73 @@ std::string BaseQuasarServer::getApplicationPath() const
     return std::string(serverSettingsPath);
 }
 
+std::string BaseQuasarServer::getProcessEnvironmentVariables() const
+{
+    std::ostringstream result;
+#ifdef __linux__
+    extern char **environ;
```
pnikiel, line 211: Please add a note explaining that this is standard POSIX interface, and maybe place a link to https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html or any other relevant reference.

Server/src/BaseQuasarServer.cpp
```diff
@@ -198,6 +204,73 @@ std::string BaseQuasarServer::getApplicationPath() const
...
+* 'quasar' runs process with admin rights (or admin shell) 'quasar(elevated)'
+*/
+std::string BaseQuasarServer::getProcessOwner() const
+{
+    std::ostringstream result;
+#ifdef __linux__
+    const auto userID = getlogin();
+    result << (userID != nullptr ? userID : "unknown");
+
+    std::string effectiveUserID = "unknown";
```
pnikiel, line 249: Please clarify: in Linux we have a distinction between UID and effective UID; the functions to get these are getuid() and geteuid(). I have rough time figuring out why it's considered that getuid() (in the submitted code) is considered "effective" while the actual effective (geteuid()) is not taken into account at all?

Server/src/BaseQuasarServer.cpp
```diff
@@ -198,6 +204,73 @@ std::string BaseQuasarServer::getApplicationPath() const
...
+    auto pwuid = getpwuid(getuid());
+    if (pwuid != nullptr && pwuid->pw_name != nullptr)
+    {
```
pnikiel, line 252: Asking for improvement: while information coming from getuid() and geteuid() is always available (reference: "man getuid") while the numeric user ID might be not mappable to its username. It actually is a common situation in distributed systems. Please improve it such that the numeric UID + numeric EUID is always printed, while respective usernames (for both UID or EUID) are printed *in addition*, if they can be obtained.

Server/src/BaseQuasarServer.cpp
```diff
@@ -284,9 +359,18 @@ int BaseQuasarServer::initializeEnvironment()
     const int ret = 0;
 #endif    
     initializeLogIt();
+    logEnvironment();
```
pnikiel, line 362: That's what I meant above -- why we couldn't just print command line args directly after initializing LogIt? Why to store it?

Server/src/BaseQuasarServer.cpp
```diff
@@ -284,9 +359,18 @@ int BaseQuasarServer::initializeEnvironment()
...
     CalculatedVariables::Engine::initialize();
     return ret;
 }
+void BaseQuasarServer::logEnvironment() const
+{
+    LOG(Log::INF) << __FUNCTION__ << std::endl << \
+        "\t Command line args: "<< boost::algorithm::join(m_commandLineArgs, " ") << std::endl << \
+        "\t Current working directory: " << boost::filesystem::current_path() << std::endl << \
```
pnikiel, line 370: A comment on filesystem was already placed.

FrameworkInternals/BoostSetup.cmake
```diff
@@ -23,7 +23,7 @@
 # and optionally added ones for the project via ADDITIONAL_BOOST_LIBS
 #
 set(Boost_NO_BOOST_CMAKE ON) # workaround for boost-1.7.0 cmake config modules: disabling search for boost-cmake to use FindBoost instead
-find_package(Boost REQUIRED regex chrono program_options thread system ${ADDITIONAL_BOOST_LIBS} )
+find_package(Boost REQUIRED regex chrono program_options thread system filesystem ${ADDITIONAL_BOOST_LIBS} )
```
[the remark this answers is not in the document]
ben-farnham: 100% standard code - do you mean `std::filesystem::current_path`. Is >=C++17 no? 

If there's some other standard code to provide same, what is it?

Server/include/BaseQuasarServer.h
ben-farnham: I did waver about this. Facts is we cannot log command line args until after LogIt is initialised. So args need to be kept around somehow.
_Could_ pass args down through method parameter lists until logEnvironment is called (after LogIt initialised), but that felt cheesy too: requires extending parameter list of a few methods.
Happy to go via the parameter list route if you think that's a better way to go.

Server/src/BaseQuasarServer.cpp
ben-farnham: Good idea, will do

Server/src/BaseQuasarServer.cpp
ben-farnham: Yep, well spotted: will change getuid to geteuid for effective account.

== for info ==
Sadly, for logged in user, getuid is not the guy (returns root if run as sudo).
`void tryThings()
{ 
  std::ostringstream result;
  
  auto getUserName = [](const uid_t& uid){
    const auto pswd = getpwuid(uid);
    const std::string name(pswd != nullptr && pswd->pw_name != nullptr ? pswd->pw_name : "unknown");
    std::ostringstream s;
    s << uid << ":" << name;
    return s.str();
  };
  
  result << getUserName(getuid()) << "(" << getUserName(geteuid()) << ")";
  std::cout << __FUNCTION__ << " " <<result.str() << std::endl;
}`

1. _bash-4.2$ ./process_owner_
`tryThings 56850:bfarnham(56850:bfarnham)`
Great!

2. _bash-4.2$ sudo ./process_owner_
`tryThings 0:root(0:root)`
From debugging/diagnosing POV I'd like to know the logged in user as well as effective process owner.
Which seems less useful than the info obtained from the current implementation, where we report on the logged in user (bfarnham) and the

Server/src/BaseQuasarServer.cpp
ben-farnham: done

Server/src/BaseQuasarServer.cpp
ben-farnham: Ah, irritatingly not so straightforward as I thought, see [this comment](https://github.com/quasar-team/quasar/pull/232#discussion_r522852521) regarding getuid and geteuid actually return. 

I mean `getlogin()`  (returns string) does not refer to same user as returned by `getuid()`, as was my (naive) hope.

And just for general hilarity, stay well clear of method `cuserid` apparently, man cuserid... 
"Nobody knows precisely what cuserid() does; avoid it in portable programs.  Or avoid it altogether"

ben-farnham: Update: added (numerical) result of geteuid to the output string. So now looks like

* `Process owner: bfarnham(56850:bfarnham)` (run normal privs)
* `Process owner: bfarnham(0:root)` (run as sudo)

Server/src/BaseQuasarServer.cpp
pnikiel: Hi @ben-farnham ,
I still think that using getlogin + geteuid is not enough. 
We should have output of either getuid + geteuid, or if you prefer getlogin + getuid + geteuid.

Reasons are that getlogin is weak. On the other hand info getuid and geteuid comes directly from the kernel.

Read this: 

>      Unfortunately, it is often rather easy to fool getlogin().  Sometimes it does not work at all, because some program > messed up the utmp file.  Often, it gives only the first 8 characters of the login
>       name.  The user currently logged in on the controlling terminal of our program need not be the user who started > it.  Avoid getlogin() for security-related purposes.

So in a world where you have: uid, euid, SUID executable bit, caps, distirbuted permissions and ACLs I think uid + euid are mandatory, getlogin() is OK to have but to me optional and might be actually misleading (see the paragraph cited above).
Note that sudo has some specific mode of operation which is different from having SUID bit set (u+sx mode); both will give you elevated rights but completely differently.

Server/src/BaseQuasarServer.cpp
pnikiel: Thanks!

Server/src/BaseQuasarServer.cpp
ben-farnham: fair enough: will publish
getlogin + uid + eiud

FrameworkInternals/BoostSetup.cmake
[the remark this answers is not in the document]
ben-farnham: OK: boost::filesystem _gone_

switching to getcwd (posix) _getcwd (windows - weirdly _getcwd is a synonym for getcwd but using that name causes deprecation error).

Server/src/BaseQuasarServer.cpp
ben-farnham: Now looks like (linux, run as sudo)
Process owner: login:bfarnham uid:0,root euid: 0,root

Server/src/BaseQuasarServer.cpp
ben-farnham: my response in comment above. SO wil close this one (leabing single place for comments)

Server/include/BaseQuasarServer.h
ben-farnham: Hi @pnikiel,
So - all other comments resolved: just this one remains. As I said, happy to change the impl to pass the info through method parameters if preferred. I felt storage option gives cleaner code, albeit at small memory price - bit no big deal either way.

BTW: you'll need to upmerge to nebula for working build - boost_filesystem gone in latest feature branch.

PR #236, opened 2021-01-08
Split between public travis yml and the non-public part (the local runner does both, the actual travis does only the public part)

pnikiel:


ben-farnham (approved):
Looks good to me

pnikiel:
thank you ;-)

parasxos:
It seems fine to me too

pnikiel:
thanks ;-)
I reverted because I accidentally pressed merge while this feat was planned for 1.5.1.

PR #241, opened 2021-02-10
add missing arg in getRestrictions()

parasxos:


ben-farnham (approved):
Yep, looks alright to me.

pnikiel:
Hi @parasxos,

while it does not crash (so, it's better) it does not deliver the expected result:

image

see, the "Value restrictions" field is empty.
I know you were only fixing something obviously broken so maybe you don't have time for full fix. If that's the case feel free to point a finger to me (I might have broken it in the meantime) or to Ben (the original translator into Jinja2).

parasxos:
Hi Piotr,

I am pointing the finger to you ;) (in a nice way)

P.

pnikiel:
Okay please merge your fix and I'll do the follow-up.
(I hope we're talking about the 2nd finger and not the 3rd finger ;-) )

parasxos:
Done. 
(of course it is the second 💅🏽)

PR #244, opened 2021-02-15
Opcua 2176 detect too old astyle

pnikiel:
Who wants a coup d'oeil on it?

CI results:
test name                                | ret code | time spent [s] 
---------------------------------------- + -------- + ---------------
open62541_default_quasar_design          |        0 |             136
open62541_test_recurrent_hasobjects      |        0 |             137
open62541_test_cache_variables           |        0 |             212
open62541_test_single_variable_node      |        0 |             137
open62541_test_methods                   |        0 |             139
open62541_test_instantiation_from_design |        0 |             124
open62541_test_config_entries            |        0 |             136
open62541_test_config_restrictions       |        0 |             136
uasdk_default_quasar_design              |        0 |              85
uasdk_test_sync_methods                  |        0 |             100
uasdk_test_async_methods                 |        0 |             111
uasdk_test_config_entries                |        0 |              86
uasdk_test_cache_variables               |        0 |             202
uasdk_test_source_variables              |        0 |             125

### DIFF
--- a/FrameworkInternals/transformDesign.py
+++ b/FrameworkInternals/transformDesign.py
@@ -108,8 +108,12 @@ def run_indent_tool(unindented_content, fout):
        fout - a file object to which the output will be written"""
     try:
         astyle_version_check_process = subprocess.run(['astyle', '--version'],
-                                                    stderr=subprocess.PIPE, check=True)
-        astyle_version_text = astyle_version_check_process.stderr.decode('utf-8')
+                                                          stderr=subprocess.PIPE,
+                                                          stdout=subprocess.PIPE,
+                                                          check=True)
+        # note to line below: whether version is printed in stdout or stderr unfortunately depends
+        # on astyle version ... ironic ;-)
+        astyle_version_text = astyle_version_check_process.stderr.decode('utf-8') + astyle_version_check_process.stdout.decode('utf-8')
         if 'Artistic Style Version 3' not in astyle_version_text:
             raise Exception(('Found astyle with incompatible version. Astyle versions prior to 3 '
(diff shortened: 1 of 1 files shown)


FrameworkInternals/transformDesign.py
```diff
@@ -107,6 +107,18 @@ def run_indent_tool(unindented_content, fout):
...
        fout - a file object to which the output will be written"""
     try:
+        astyle_version_check_process = subprocess.run(['astyle', '--version'],
+                                                          stderr=subprocess.PIPE,
+                                                          stdout=subprocess.PIPE,
+                                                          check=True)
+        # note to line below: whether version is printed in stdout or stderr unfortunately depends
+        # on astyle version ... ironic ;-)
+        astyle_version_text = astyle_version_check_process.stderr.decode('utf-8') + astyle_version_check_process.stdout.decode('utf-8')
+        if 'Artistic Style Version 3' not in astyle_version_text:
```
ben-farnham, line 117: Is this a built-in future problem ? I mean when astyle 4.0 comes out.

Maybe could regex the string for 'Artistic Style Version X', and barf if X <3 ?

pnikiel: Hi Ben,

I have no idea how the version tag will look like with 4.X so I'd prefer not to make predictions.
In fact the primary reason this is added is because CC7 is defined as our baseline platform and astyle 2.X seems still kind of present there (might be my fault as it was me who 'populated' astyle). I think that once the baseline moves on from CC7 we could get rid of that check.

parasxos (approved):

pnikiel:
Thanks. I understand we have no in-depth problems with this one.

PR #247, opened 2021-02-23
add ccache in cmakelists

parasxos:


CMakeLists.txt
```diff
@@ -23,6 +23,12 @@ cmake_minimum_required(VERSION 3.10)
 # the line below is added to quasar to keep backwards compatibility for the transition period.
 cmake_policy(SET CMP0065 OLD)
 
+find_program(CCACHE_PROGRAM ccache)
+if(CCACHE_PROGRAM)
+    message(STATUS "Found ccache, using it")
```
ben-farnham, line 28: is it possible to add the ccache status file location to this message. I mean is there a CMake related variable that could be queried and dropped in to this string ? 

Like some ${CMAKE_CCACHE_LOC} type thing

parasxos: There is the environment variable CCACHE_DIR which ,when set, specifies where ccache will keep its cached compiler output. The default is "$HOME/.ccache".

pnikiel: Hi @parasxos ,
to my mind this logical chain:

if ccache_was_found then let's_use_it

is not necessarily very welcome. CCache might by lying around but you still might not be interested to use it.

therefore I'm inviting you to consider either of these:
(1) inject your ccache via build configs,
(2) inject your ccache via ProjetSettings,
(3) supply an extension to quasar's CMakeLists with prologue similarly to the epilogues. IMO this should let people do ccache within the prologue and give control and much higher extensibility for other things that might go into such a prologue.

FYI We already have epilogues:

https://github.com/quasar-team/quasar/blob/master/CMakeLists.txt#L149

parasxos: ccache is not installed by default. If it is installed on a development machine there is a big chance it was done in order to use it. I assume, probably wrongly(?), in most if not all of the cases. Though I would like to understand exactly that point, "is not necessarily very welcome".

Could we think of an example where if ccache is installed, why someone would not like to use it?

pnikiel: Hi @parasxos ,
I think that if CCache (or a similar tool) was always supposed to be used with CMake (no matter what) then it would simply be "on" by default in CMake rather than a possible option to turn on. So somehow CMake also behaved quite conservatively there.

I think such tools require more careful evaluation. I believe it's a good compromise to put it into build configs because you could use it and we could evaluate it without possible cons that we might not be aware of yet. As you know every human-made object has pros and cons, so let's have more time to understand what possible cons are.

parasxos: I have been evaluating this for the last month without any issue and the tool has a history of 17 years. CMake has official support for it and native integration. Why it is not enabled by default, I could think of many reasons (many not technical) but I would take distance from drawing conclusions.

As a server developer, I would had a preference to have that option provided by the framework instead of copying those lines from build config to build config. 
We have in quasar Ninja which is optionally enabled. What if we get inspired by that and (1) keep it in CMakeLists and (2) use just a switch in the build configs to enable it?

Like this we get the time to evaluate it and if the team is happy with the evaluation, in the future we could just turn on the option by default in the CMakeLists, minimizing the future changes.

pnikiel: Could you clarify what you propose?
Could you write what you want to put into master CMakeLists and in some build config?

parasxos: Hi @pnikiel, 
I would take your suggestion and put it into the build config of our servers.
Thanks both for the review,
Paris

PR #266, opened 2022-01-13
getFullName works properly for single item node

pnikiel:


Configuration/templates/designToConfigurator.jinja
```diff
@@ -90,7 +91,20 @@
 
       // link address space and device side objects
       asItemChildByDesign->linkDevice( dItemChildByDesign );
-      dItemChildByDesign->linkAddressSpace( asItemChildByDesign, asItemChildByDesign->nodeId().toString().toUtf8() );
+      // establish the full name of an object.
+      {% if designInspector.is_class_single_variable_node(className) %}
+        {
+          std::string singleItemNodeFullName = asItemChildByDesign->nodeId().toString().toUtf8();
```
ben-farnham, line 97: Think it's worth making a method for this? Looks like the exact same logic  in line 97 -> 102 as 225 -> 230

ben-farnham (approved):
Looks good - just one open question in the review.

pnikiel:
Review cmnts implemented.

PR #276, opened 2022-06-21
Opcua 2757 uasdk info

parasxos:


ben-farnham:
So, for my education, what sort of format does the string returned from `VersionInfoCoreModule::getCoreModuleVersionInfo().toUtf8()`
return (that you need to replace '\' with ' ')?

parasxos:
Here is an example of what  VersionInfoCoreModule::getCoreModuleVersionInfo().toUtf8() will return: "1.6.5.478\Nov 15 2021 18:11:04 GNU\C++ 11.1". 
In the ticket https://its.cern.ch/jira/browse/OPCUA-2757, there is an example of how the result looks like.

ben-farnham:
Ah, thanks.

and in that case - I like it!

ben-farnham (approved):

PR #282, opened 2022-07-21
Opcua 2768 propagate is key into unique constraint

pnikiel:


ben-farnham (approved):
general comment/question. Like the idea, seems good

Just wondering... seems every time the writeUniqueConstraintPerHasObject macro is invoked there is somewhat similar looking wrapping (xsd and jinja stuff).

A question: is there scope to remove/reduce that similar looking code by either making the macro act on a hasObjects list (so arg list different) or making an outer macro that acts on hasObjects list ?

pnikiel:
Thanks @ben-farnham, indeed you could reduce these 3 lines (or 6) lines if you wrap it up altogether. I however see that it makes it less readable and less obvious what that is supposed to do, so for the moment I'd prefer not to do that optimization.

PR #286, opened 2022-08-03
Opcua 2788 test for default instance name

pnikiel:


ben-farnham (approved):
Looks good.

For my own education... what is actually tested here? I guess it's more than just 'server builds' since there is a config file. Is it 'server builds and runs without error' ?

pnikiel:
Hi @ben-farnham 
indeed it tests that a server builds and runs with the specified configuration.
With a very small investment we can also make sure that the address-space contents as is supposed (e.g. the TestClass at address "ns=2;s=XX").

PR #298, opened 2022-11-21
Refusing the OP instead of throwing up

pnikiel:


parasxos (approved):
Everything is better than "throwing up"

PR #303, opened 2023-01-31
Opcua 2962

parasxos:


FrameworkInternals/parser_module.py
```diff
@@ -0,0 +1,193 @@
...
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS  INTERRUPTION) HOWEVER CAUSED AND ON
+ANY  THEORY  OF  LIABILITY,   WHETHER IN  CONTRACT, STRICT  LIABILITY,  OR  TORT
+(INCLUDING  NEGLIGENCE OR OTHERWISE)  ARISING IN ANY WAY OUT OF  THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+@contact:    quasar-developers@cern.ch
+'''
+
+import argparse
+
+class MyFormatter(argparse.HelpFormatter):
```
pnikiel, line 34: I'd suggest something like QuasarArgsHelpFormatter or something like it.

FrameworkInternals/parser_module.py
```diff
@@ -0,0 +1,193 @@
...
+
+            # add the item to the list
+            self._add_item(self._format_action, [action])
+
+def quasar_parser():
+    """
+    Returns the parser for the quasar command line interface
+    """
+
+    def formatter_class(prog): return MyFormatter(prog, max_help_position=30, width=150)
```
pnikiel, line 67: Though extreme cosmetics, I'd stick more to PEP-8 here: put return in the next line.

pnikiel (commented):
As passed to the author personally, I could have only had a quick look at the code because this looks to me as the 1st step in a longer series of steps of transition of the earlier CLI into argaprse's one; without understanding the full planned work I can't really say if that is a change that would make it easier to maintain overall and make it more robust (thus, whether in my opinion it'd be beneficial to quasar or not).

I do not see any obvious problems with it nevertheless I think that some solid testing is necessary, especially that quasar has many intricate commands.

I believe that the code should be run through pylint as I believe that in some places (especially the new file parser_module.py) PEP-8 could be improved for instance regarding whitespace before and after the '=' sign.

PR #318, opened 2023-03-24
Opcua 3008

parasxos:


ben-farnham (commented):
Hi Paris, I think it looks good, the only comment I have is it lacks documentation no? What is a sanitizer for ? How do I activate ? How is information about unsanitary code reported ?

Also, I think it should be explicitly documented that (I assume) this is linux only. Poor old windows users should be informed at least.

PR #329, opened 2024-10-23
modified template to create compiling code in the case that a devicel…

ben-farnham:
…ogic class has no valid devicelogic parent

### DIFF
--- a/Device/templates/designToDeviceBaseHeader.jinja
+++ b/Device/templates/designToDeviceBaseHeader.jinja
@@ -21,10 +21,10 @@
 
 {% macro get_parent_struct(class_name) %}
   {% set parent_class_name = designInspector.get_parent(class_name) %}
-  {%- if parent_class_name == None or not designInspector.class_has_legit_device_parent(parent_class_name) -%}
-    struct{/*No Device Logic for parent {{parent_class_name}} of {{class_name}}*/}
-  {%- else -%}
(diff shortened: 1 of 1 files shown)


ben-farnham:
Hey folks, this is my proposed fix for [OPCUA-3228].

Which is a blocker for the implementation needed to fix [[OPCUA-3226](https://its.cern.ch/jira/browse/OPCUA-3226)]

FrameworkInternals/DesignInspector.py
```diff
@@ -114,9 +114,7 @@ def class_has_legit_device_parent(self, class_name):
             return False # referenced from multiple parents, so far sure no unique parent
         if has_objects_origins[0] == "Root":
             return True # Root by definition has device logic
-        if not self.class_has_device_logic(has_objects_origins[0]):
-            return False
-        return True
+        return self.class_has_device_logic(has_objects_origins[0])
```
parasxos, line 117: Let’s keep the explicit return True and return False statements for clarity and to ensure consistent boolean results, which makes the code more maintainable and robust against future changes.

parasxos (changes requested):

FrameworkInternals/DesignInspector.py
ben-farnham: OK, reverted DesignInspector, now as master

Inverted if/else logic in template

Should be good to go, have a look and see what you think

PR #330, opened 2024-10-23
Opcua 3226 initialise source variable thread pool from smd

ben-farnham:
Proposed fix for OPCUA-3226 (SMD refactor broke source variable thread pool initialisation from config.xml)

**Note** this fix requires PR#331 to be merged to master as pre-requisite. Required since the simplest fix for this issue requires (in meta's Design.xml) creating a devicelogic class that does not have a unique parent class also having devicelogic.


parasxos (changes requested):
files.txt is a generated file thus should stay untouched. It will recreated with the next RC release.

Other than that, no other comment. 

Thanks Ben

PR #333, opened 2025-04-29
Handle empty OPCUA_TOOLKIT_LIBS case in CMakeLists.txt with condition…

parasxos:
…al logic.


ben-farnham (changes requested):
I think it's possible to do this in a one liner with
`string(REPLACE "-l" "_l" OPCUA_TOOLKIT_LIBS_ESCAPED "${OPCUA_TOOLKIT_LIBS})"`
i.e. just adding the double quotes around the last arg in the call above.

Then if OPCUA_TOOLKIT_LIBS is undefined the string REPLACE call will act on an empty string and return an empty string as result.

What you think? Who doesn't love a bit of cmake golf

PR #348, opened 2026-04-13
OPCUA-3326 Modernize CMake 3.26 minimum to Ensure Windows and Linux Parity

luismiguensfernandez:
## CMakeLists.txt
The core of the changes includes using find_package for everything needed and trying to be closer to CMake recommendations. Tested in Windows 2025 (Visual Studio 2026) and Linux (Alma 10).
Changes: (the most important in my opinion)

- custom build config is optional (`run_test_case.py` does not set the build config anymore), not mandatory. All tests and OPC Servers of TMA compile without custom config on Windows and Linux
- auto-select backend based on open6-compat enable/disabled
- enable all available backends (compatible with the context). Ex. If UASDK is enabled, then UATRACE for LogIt is enabled, If in Windows, WINDOWS_DEBUGGER enabled.

## Header of OptionalModule.cmake to cmake_minimum_required(VERSION 3.26)
- Only bump version.

## BoostSetup.cmake
- System was removed as component
- Rely on find_package(CONFIG) rather than FindBoost
- FindBoost is deprecated (CMake 3.30). Starting Boost 1.70, find_package(CONFIG) is the preferred way (supported in CMake 3.26 too)
Since 1.69, Boost.System is header-only. Alma 10 currently has 1.83.0-5. Windows build needs Boost 1.90. The system does not provide a dummy component for System anymore, so it must go away in the cmake file.

References:

https://www.boost.org/releases/1.69.0/
https://gitlab.kitware.com/cmake/cmake/-/work_items/19402
https://github.com/msys2/MINGW-packages/issues/22081

## Discovery UA-SDK tested for 1.8.9
- Auto discovers UA-SDK 1.8.9. Users will custom location can set the environment variable `UNIFIED_AUTOMATION_HOME` or add it to the `CMAKE_PREFIX_PATH` via custom build.

## Impact on existing projects

1. A user using devel dnf packages, or installed in default locations (/usr, /usr/local), or using the right environment variables (ex. inside Docker):
- Delete the custom build config should be enough

2. A user with custom requirements:
Example:

Has UA SDK and Boost in their home folder
Want to disable the BOOST backend of LogIt
Want to compile with UASDK, even though the project has enabled open6-compat, therefore disabling auto-selection.
The custom config should look like:

```
set(LOGIT_BACKEND_BOOSTLOG OFF)
set(BACKEND_UATOOLKIT ON)
list(PREPEND CMAKE_PREFIX_PATH "/home/hancock/workspace/boost/"
                               "/home/hancock/workspace/uasdk/")
```

**CMake version (as today) for reference:**

Current CMake Windows 2025: CMake 4.3.1

Alma 10: CMake 3.30.5

Alma 9: 3.26.5

### DIFF
--- a/.CI/run_test_case.py
+++ b/.CI/run_test_case.py
@@ -20,9 +20,8 @@ def clone_quasar(test_branch):
 def prepare_opcua_backend(opcua_backend, open62541_compat_branch):
     if opcua_backend == 'o6':
         invoke_and_check(f'./quasar.py enable_module open62541-compat {open62541_compat_branch}')
-        invoke_and_check(f'./quasar.py set_build_config open62541_config.cmake')
     elif opcua_backend == 'uasdk':
-        invoke_and_check('./quasar.py set_build_config .CI/travis/build_configs/uasdk-eval.cmake')
+        pass
     else:
         raise Exception(f"OPCUA backend {opcua_backend} was not recognized")
 
--- a/.CI/run_test_case.py
+++ b/.CI/run_test_case.py
@@ -20,8 +20,9 @@ def clone_quasar(test_branch):
 def prepare_opcua_backend(opcua_backend, open62541_compat_branch):
     if opcua_backend == 'o6':
         invoke_and_check(f'./quasar.py enable_module open62541-compat {open62541_compat_branch}')
+        invoke_and_check(f'./quasar.py set_build_config open62541_config.cmake')
     elif opcua_backend == 'uasdk':
-        pass
+        invoke_and_check('./quasar.py set_build_config .CI/travis/build_configs/uasdk-eval.cmake')
     else:
         raise Exception(f"OPCUA backend {opcua_backend} was not recognized")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,7 @@ set(TARGET_LIBS
         ${SERVER_LINK_LIBRARIES}
         ${OPCUA_TOOLKIT_LIBS}
         ${XML_LIBS}
+        ${QUASAR_SERVER_LIBS}
         Threads::Threads
         )
 
@@ -246,8 +247,6 @@ if(ENABLE_QUASAR_SANITIZERS)
     include (${PROJECT_SOURCE_DIR}/cmake/QuasarSanitizers.cmake)
     enable_quasar_sanitizers(${EXECUTABLE})
 endif()
-
-message(STATUS "Target Link Libraries: ${TARGET_LIBS}")
 
 target_link_libraries ( ${EXECUTABLE} PRIVATE ${TARGET_LIBS} )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ endif()
 if(EXISTS ${PROJECT_SOURCE_DIR}/FrameworkInternals/build_config_selector.cmake)
 	  include( ${PROJECT_SOURCE_DIR}/FrameworkInternals/build_config_selector.cmake )
 else()
-     message(STATUS "Extra build configuration has not been chosen.")
+     message(FATAL_ERROR "Build configuration has not been chosen yet. Please run 'quasar.py set_build_config'")
 endif()
 
 include( ProjectSettings.cmake )
@@ -76,14 +76,10 @@ find_package(LibXml2 CONFIG REQUIRED)
 message(STATUS "Found libxml2: ${LIBXML2_INCLUDE_DIRS}")
 list(APPEND QUASAR_GLOBAL_INCLUDE_DIRECTORIES ${LIBXML2_INCLUDE_DIRS})
 
-# Backward compatibility for projects generated in the past.
-# DO NOT REMOVE unless all legacy CMake code has stopped using XML_LIBS.
-if(NOT DEFINED XML_LIBS OR XML_LIBS STREQUAL "")
-  set(XML_LIBS
-      XercesC::XercesC
-      LibXml2::LibXml2
-  )
-endif()
+list(APPEND XML_LIBS
+    XercesC::XercesC
+    LibXml2::LibXml2
+)
 
 if(${BUILD_SERVER_SHARED_LIB})
 	list(APPEND QUASAR_GLOBAL_COMPILE_DEFINITIONS BUILDING_SHARED_OBJECT)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,18 +43,6 @@ include( ProjectSettings.cmake )
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
-if(NOT DEFINED LOGIT_BACKEND_STDOUTLOG)
-    set(LOGIT_BACKEND_STDOUTLOG ON)
-endif()
-
-if(NOT DEFINED LOGIT_BACKEND_BOOSTLOG)
-    set(LOGIT_BACKEND_BOOSTLOG ON)
-endif()
-
-if(WIN32 AND NOT DEFINED LOGIT_BACKEND_WINDOWS_DEBUGGER)
-    set(LOGIT_BACKEND_WINDOWS_DEBUGGER ON)
-endif()
-
 if(NOT ${PROJECT_SOURCE_DIR} STREQUAL ${PROJECT_BINARY_DIR})
 message("Building:")
 message("PROJECT_SOURCE_DIR= ${PROJECT_SOURCE_DIR}")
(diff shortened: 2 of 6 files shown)


cmake/FindOpcUaToolkit.cmake
```diff
@@ -16,212 +16,150 @@
 #
 # FindOpcUaToolkit.cmake
```
parasxos, line 17: The original authorship and the usage documentation block got dropped here. Please put them back and add yourself.

CMakeLists.txt
```diff
@@ -19,73 +19,146 @@
...
+include( ProjectSettings.cmake )
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
+if(NOT DEFINED LOGIT_BACKEND_STDOUTLOG)
+    set(LOGIT_BACKEND_STDOUTLOG ON)
+endif()
+
+if(NOT DEFINED LOGIT_BACKEND_BOOSTLOG)
+    set(LOGIT_BACKEND_BOOSTLOG ON)
```
parasxos: These defaults have no effect. LogIt calls option() on these variables in its own CMakeLists and CMP0077 is not set, so option() clears the normal variable back to its own default (OFF). Confirmed across four ATLAS servers in testing. The fix would be cmake_policy(SET CMP0077 NEW), but I'd rather you take this out of the PR entirely since it's LogIt framework config, not Windows scope.

CMakeLists.txt
```diff
@@ -19,73 +19,146 @@
...
+
+# It applies to everything, build config and project settings inclusive
+if(WIN32 AND MSVC)
+    add_compile_definitions(NOMINMAX WIN32_LEAN_AND_MEAN NOGDI) # Prevent collisions due to windows.h
+endif()
+
+if(EXISTS ${PROJECT_SOURCE_DIR}/FrameworkInternals/build_config_selector.cmake)
+	  include( ${PROJECT_SOURCE_DIR}/FrameworkInternals/build_config_selector.cmake )
+else()
+     message(STATUS "Extra build configuration has not been chosen.")
```
parasxos: Hard no on this one. Changing whether set_build_config is mandatory is a change to how quasar works, not a Windows parity fix. Revert to FATAL_ERROR. If you want to propose making build config optional that's a separate ticket.

CMakeLists.txt
```diff
@@ -157,13 +266,19 @@ if(ENABLE_QUASAR_SANITIZERS)
     enable_quasar_sanitizers(${EXECUTABLE})
 endif()
 
-target_link_libraries ( ${EXECUTABLE} ${TARGET_LIBS} )
+message(STATUS "Target Link Libraries: ${TARGET_LIBS}")
```
parasxos: Debug output, please remove.

CMakeLists.txt
```diff
@@ -132,22 +225,38 @@ add_custom_command(
...
+endif()
+
+target_link_libraries(${EXECUTABLE} PRIVATE quasar_project_options)
+if(QUASAR_TARGET_LINK_DIRECTORIES)
+    target_link_directories(${EXECUTABLE} PRIVATE ${QUASAR_TARGET_LINK_DIRECTORIES})
+endif()
+
+find_package(Threads REQUIRED)
 
 set(TARGET_LIBS
```
parasxos, line 235: Hard no. QUASAR_SERVER_LIBS is consumed by existing build configs including our own open62541_config.cmake. Dropping it silently breaks downstream projects. Backwards incompatibility is not on the table here. Put it back in TARGET_LIBS.

FrameworkInternals/BoostSetup.cmake
```diff
@@ -22,12 +22,28 @@
...
-find_package(Boost REQUIRED regex chrono program_options thread system ${ADDITIONAL_BOOST_LIBS} )
+if (DEFINED ENV{BOOST_HOME} AND NOT "$ENV{BOOST_HOME}" STREQUAL "")
+    set(BOOST_HOME "$ENV{BOOST_HOME}")
+    list(PREPEND CMAKE_PREFIX_PATH "${BOOST_HOME}")
+    set(Boost_USE_STATIC_LIBS ON)
+    message(STATUS "Using BOOST_HOME environment variable: [${BOOST_HOME}]")
+    message(STATUS "This will set BOOST to that folder in static mode")
+endif()
+
+set(_QUASAR_BOOST_COMPONENTS regex chrono program_options thread log ${ADDITIONAL_BOOST_LIBS})
```
parasxos: Adding log here makes Boost.Log a mandatory dep for every quasar project. Previously it was only pulled in when LOGIT_BACKEND_BOOSTLOG was on, which was off by default. This is a new framework wide requirement, not a Windows need. Should come out of this PR.

.CI/run_test_case.py
```diff
@@ -20,9 +20,8 @@ def clone_quasar(test_branch):
 def prepare_opcua_backend(opcua_backend, open62541_compat_branch):
     if opcua_backend == 'o6':
         invoke_and_check(f'./quasar.py enable_module open62541-compat {open62541_compat_branch}')
-        invoke_and_check(f'./quasar.py set_build_config open62541_config.cmake')
     elif opcua_backend == 'uasdk':
-        invoke_and_check('./quasar.py set_build_config .CI/travis/build_configs/uasdk-eval.cmake')
+        pass
```
parasxos: What's the user need driving the removal of set_build_config from CI? This is CI infrastructure and I'd rather handle it separately once the find_package changes are properly evaluated. Also bare pass with no comment is not great style.

parasxos (changes requested):
Luis, thanks for the work on this. I want to be direct about how I'm reading it, because I think we should get the process right before we dig into the code.

The scoping agreement for OPCUA-3326 was focused: FindOpcUaToolkit for UASDK 1.8.x and the Windows/Linux parity fixes. What I'm seeing is broader than that. Boost setup rewrite, backend auto-selection, dependency resolution via find_package, LogIt defaults, build config becoming optional, CI changes. Some of these are fine ideas individually but they're framework level decisions, not Windows or LogIt scope, and I'd rather see them as request tickets so I can triage and sequence them properly.

What belongs in this PR is the Windows parity work and FindOpcUaToolkit. The MSVC output dir trick, the crypt32 link, the WIN32_LEAN_AND_MEAN, the Windows debugger logit backend, all good. FindOpcUaToolkit has real bugs though and I'll inline on those.

For everything else, could you open separate tickets describing what each change enables for a user and I'll go through them. Modernization by itself is not a strong enough driver, I need the user need behind each one.

I also ran the PR against four ATLAS servers in our Alma 9 CI image and found a couple of things I'll flag inline. Will re-review once the scope is trimmed and the FindOpcUaToolkit fixes are in.

.CI/run_test_case.py
luismiguensfernandez: Comment addressed

FrameworkInternals/BoostSetup.cmake
luismiguensfernandez: Comment addressed

CMakeLists.txt
luismiguensfernandez: Comment addressed

CMakeLists.txt
luismiguensfernandez: Comment addressed

CMakeLists.txt
luismiguensfernandez: Comment addressed

CMakeLists.txt
luismiguensfernandez: Not a problem in recent LogIt (because the minimum_required bump includes that policy), but removing everything LogIt-related from this MR.

cmake/FindOpcUaToolkit.cmake
luismiguensfernandez: Comment addressed

parasxos:
Luis, thanks for the quick fixes on the inline items. Re-ran the four ATLAS servers and the UASDK ones are clean, appreciate it.

I went back and reconciled the PR scope against what was agreed on 31.03.2025 in R1 and what ended up in OPCUA-3326. There's a gap. The oral agreement that day was Windows parity plus FindOpcUaToolkit for UASDK 1.8.x, and Ben said on the same topic "I think it could be made smaller." The ticket you wrote afterward went broader than that, and the PR went broader than the ticket. The ticket itself was never formally approved. It's still Open.

So I need to bring this back in line. For OPCUA-3326 the scope is Windows parity in CMakeLists.txt (the MSVC output dir, the Windows compile defs, crypt32, WIN32 guards), the FindOpcUaToolkit.cmake rewrite for 1.8.x with the lib-search and version-detection issues fixed, and the OptionalModule.cmake version bump. That's it.

Everything else comes out of this PR. cmake_minimum_required bump to 3.26 in root, find_package modernization for XercesC OpenSSL LibXml2 Python3 Threads, BoostSetup CONFIG mode with BOOST_HOME, backend auto-selection, quasar_project_options with target_link_libraries PRIVATE, CMAKE_CXX_STANDARD 17 centralization, SERVERCONFIG_LOADER move. Each of these is a framework-level change that needs its own evaluation and its own user-need justification. Some I'll likely agree with. Others have real technical risk we haven't explored.

I'll open the follow-up tickets myself this week so you don't have to. You'll get to fill in the user-need-and-motivation field for the ones you want to push, and I'll triage from there.

Easiest path from here is probably to close this PR and open a fresh one with just the tight scope. Keeps the diff clean and avoids confusion about what's in. Let me know what works for you.

luismiguensfernandez:
Replaced by https://github.com/quasar-team/quasar/pull/352

PR #352, opened 2026-04-27
OPCUA-3326 Modernize CMake 3.26 minimum to Ensure Windows and Linux Parity

luismiguensfernandez:
Summary of changes

### AddressSpace/templates/designToGeneratedCmakeAddressSpace.jinja
- Use `${PYTHON_COMMAND}` (which was hardcoded to python3, although line 35 was already using ${PYTHON_COMMAND}), without it Windows cannot go through the config step

### CMakeLists.txt
Most important changes:
- cmake_minimum_required
- if(WIN32 AND MSVC) part, indicating CMake to prefer CONFIG over MODULE in Windows (bug in LibXml2 module for Windows)
- Logic reading environment variables (if they exist), prepending to search path and find_package; appending result to existing variable (i.e XML_LIBS)
- add_definitions(-DBACKEND_OPEN62541) and include, the same as existing BACKEND_UATOOLKIT and unsure why it was not like that already.

### BoostSetup and related
- Almost identical to master regarding existing variables to ignore boost of set it manually, adding BOOST_HOME and forcing CONFIG mode for find_package

### FrameworkInternals/OptionalModule.cmake
- Bump cmake minimum

The tests of Quasar are passing in my machine, and I tested on Wiener (both Linux/Windows). Tell me when it is ok and I will open the pull-request.


CMakeLists.txt
```diff
@@ -71,7 +100,9 @@ else()
     include_directories(${BOOST_PATH_HEADERS})  # this falls back to the old solution when we didn't yet use find_package(Boost)
 endif()
 
-set(PYTHON_COMMAND python3)
+find_package(Python3 COMPONENTS Interpreter REQUIRED)
+set(PYTHON_COMMAND Python3::Interpreter)
```
parasxos: Technical question on this line. Using the imported target Python3::Interpreter works in most COMMAND contexts because CMake auto-resolves it, but it has edge cases in string interpolation and in some custom command patterns. The more portable pattern is the explicit executable path, as you had in the earlier iteration:

set(PYTHON_COMMAND "${Python3_EXECUTABLE}")

Happy to keep the target form if you have a specific reason for it, just let me know.

FrameworkInternals/BoostSetup.cmake
```diff
@@ -22,12 +22,26 @@
 # automatically set up boost with libs needed by quasar
 # and optionally added ones for the project via ADDITIONAL_BOOST_LIBS
 #
-set(Boost_NO_BOOST_CMAKE ON) # workaround for boost-1.7.0 cmake config modules: disabling search for boost-cmake to use FindBoost instead
-find_package(Boost REQUIRED regex chrono program_options thread system ${ADDITIONAL_BOOST_LIBS} )
+if(DEFINED ENV{BOOST_HOME} AND NOT "$ENV{BOOST_HOME}" STREQUAL "")
+    set(BOOST_HOME "$ENV{BOOST_HOME}")
+    list(PREPEND CMAKE_PREFIX_PATH "${BOOST_HOME}")
+    set(Boost_USE_STATIC_LIBS ON)
```
parasxos: Technical question here. This forces static Boost linking whenever BOOST_HOME is set. In a mixed-linking scenario static Boost can cause ABI surprises if the rest of the project links some libraries dynamically. Could you share the reason for forcing static, or would it make sense to gate it behind an explicit opt-in flag like BOOST_USE_STATIC rather than implying it from BOOST_HOME? If there is a specific setup that requires static, we can document that and keep the current behaviour, just want to make sure it is an intentional choice.

luismiguensfernandez: In TMA we link BOOST statically (Windows and Linux), as well as other dependencies (ex. UA-SDK, libxml, etc). In most cases, `find_package` will return either the dynamic library or static (in that order of preference) if found, but in Boost CONFIG the situation does not follow the convention of the rest of the dependencies, and it will always try to find the dynamic version, rejecting the static version unless you set  `Boost_USE_STATIC_LIBS `.

As the dynamic version is usually in `/usr/lib64` or similar and already by default in the search path of `find_package`, there is no need to set the `BOOST_HOME`, but maybe we could make it more explicit, such as:

`BOOST_STATIC_HOME`: Add to the path and set `Boost_USE_STATIC_LIBS `
`BOOST_HOME`: Only add to the path, therefore dynamically-only behavior, for the people who have Boost in non-standard directories.

CMakeLists.txt
luismiguensfernandez: Changed to: `set(PYTHON_COMMAND "${Python3_EXECUTABLE}")`

FrameworkInternals/BoostSetup.cmake
parasxos: Thanks Luis. Note BoostConfig defaults to static on Windows already. Could we use BOOST_ROOT (the CMake standard) instead of the split, and let users set Boost_USE_STATIC_LIBS themselves when needed? One variable, location and linkage stay orthogonal.

luismiguensfernandez: Done. https://github.com/quasar-team/quasar/pull/352/changes/b009711c974e75d0fe1f3088a3fb88fcac4a52d5

parasxos (approved):
Thanks Luis. Both points addressed cleanly. Approved.

PR #356, opened 2026-06-03
OPCUA-3366: Make code generation write-if-different (stop spurious rebuilds)

parasxos:
## What
Makes quasar's code generation **write-if-different**: a build with an unchanged `Design.xml` no longer rewrites byte-identical generated files with a fresh mtime, so CMake/make stop rebuilding everything (OPCUA-3366). Turns the `framework_invariants` job (currently red on master) green.

## The fix — +49/-16, 4 files, one shared helper
- **`transformDesign.py`** — new `write_if_different(path, bytes)` (binary compare → Windows-safe; first generation writes); `run_indent_tool` now *returns* the indented bytes; `transformDesignByJinja` compares the **final** (post-astyle) bytes before writing.
- **`merge_design_and_meta.py`** — `write_to_file` takes a path and write-if-differents (the merge must own its destination to compare *before* truncating); legacy pre-opened-handle form kept for the standalone `process_args` path.
- **`generateCmake.py`** — pass the merge destination as a path (1 line).
- **`manage_files.py`** — adapt `_style_it` to `run_indent_tool`'s new return value (1 line).

Generated **content is unchanged** — only redundant writes are skipped.

## Testing
- **Flips the landed test green:** `.CI/test_no_spurious_rebuild.py` (red on master) now passes; `.CI/test_upgrade_override.py` still passes.
- **Byte-identical content:** 7 representative outputs (astyle C++, plain XML/cmake, the merge) are byte-for-byte identical with vs without the fix.
- **Cross-distro hermetic:** green on AlmaLinux 9 (py3.9), AlmaLinux 10 (py3.12), Ubuntu 24.04 (py3.12).
- **Real end-to-end (alma9 toolchain container):** full `quasar.py build` succeeds; the server's exported NodeSet2 **matches the golden reference**; build → rebuild with no change → **0 Design-derived recompiles** (only env-derived `MetaBuildInfo` rebuilds, which is correct), 0 generated sources rewritten.
- Edge cases: first-gen writes, identical skips (mtime kept), changed rewrites, CRLF/NUL binary-exact, partial-write cleanup preserved, all 5 callers safe.

Refs OPCUA-3366.

### DIFF
--- a/FrameworkInternals/generateCmake.py
+++ b/FrameworkInternals/generateCmake.py
@@ -68,7 +68,7 @@ def generateCmake(context, *args):
     merge_user_and_meta_design(
         user_file   = open(mode='r', encoding='utf-8', file=os.path.join(projectSourceDir, 'Design', 'Design.xml')),
         meta_file   = open(mode='r', encoding='utf-8', file=os.path.join(projectSourceDir, 'Meta',   'design', 'meta-design.xml')),
-        merged_file = open(mode='w', encoding='utf-8', file=os.path.join(projectBinaryDir, 'Design', 'DesignWithMeta.xml'))
+        merged_file = os.path.join(projectBinaryDir, 'Design', 'DesignWithMeta.xml')
     )
 
     transformByKey(TransformKeys.AS_CMAKE,     {'context':context})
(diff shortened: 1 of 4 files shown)


jsouter (approved):
Tested with CanOpenOpcUa, works as expected

PR #357, opened 2026-06-03
OPCUA-3367: Make device_report type-aware (catch silently-unbound delegates)

parasxos:
## What
Makes `device_report` **type-aware** so it catches a delegated cache-variable delegate whose `D<Class>` parameter type has drifted from the design — a *silently unbound* delegate (OPCUA-3367). Flips the landed `.CI/test_no_unbound_delegate.py` green.

## The bug (reproduced)
A delegated CV's method signature embeds its data type. Change the type and the frozen, skip-if-exists `D<Class>` delegate keeps the old type. `override` turns that into a compile error in the standard path — but a legacy/hand-written delegate without `override` (or one a user de-`override`s to clear a red build) **drifts silently**: it builds, the server runs, the delegate binds to the old type, and `device_report` — matching by method *name* only — reported it identically to a correct one.

## The fix — +35/−2, one file
- New read-only helper `_get_expected_delegate_param_types` → `{write<Cv>: expected C++ param type}`, mirroring the `writeCacheVarSig` macro exactly (scalar `<dataType>`, array `std::vector<<dataType>>`).
- `deviceReport` captures the declared param list and adds a distinct **`delegate type mismatch`** verdict on drift.
- **`_get_expected_virtual_methods` is untouched** → `addOverrideToDeviceClasses` unaffected (`test_upgrade_override.py` still passes). No new files/modules, no runtime/`Base_D` change. Matching is whitespace/namespace-tolerant and boundary-anchored.

## Testing
- **Flips `.CI/test_no_unbound_delegate.py` red→green**; `.CI/test_upgrade_override.py` still green.
- **10/10 false-positive guard matrix**: correct delegates (scalar of any type, `::`-qualified, array, spaced `std::vector< T >`) never flagged; real drift (type change, `Int16←Int32`, array↔scalar) flagged.
- **Cross-distro on Docker**: green on AlmaLinux 9 (py3.9 — the CI image), AlmaLinux 10 (py3.12), Ubuntu 24.04 (py3.12).

Refs OPCUA-3367.

### DIFF
--- a/.CI/test_no_unbound_delegate.py
+++ b/.CI/test_no_unbound_delegate.py
@@ -51,6 +51,9 @@ _DESIGN = '''<?xml version="1.0" encoding="UTF-8"?>
     <d:cachevariable name="reading" dataType="{dtype}" addressSpaceWrite="delegated"
        initializeWith="valueAndStatus" initialStatus="OpcUa_BadWaitingForInitialData"
        initialValue="0" nullPolicy="nullForbidden"/>
+    <d:sourcevariable name="sample" dataType="{dtype}"
+       addressSpaceRead="synchronous" addressSpaceWrite="synchronous"
+       addressSpaceReadUseMutex="no" addressSpaceWriteUseMutex="no"/>
   </d:class>
   <d:root><d:hasobjects instantiateUsing="configuration" class="Sensor"/></d:root>
 </d:design>'''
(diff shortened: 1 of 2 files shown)


jsouter (dismissed):
Upgraded CanOpenOpcUa to this branch and satisfied that changing the dataType of a delegated cache variable the report is now reported by device_report

PR #361, opened 2026-06-09
OPCUA-3341 Add clang + open62541 CI job

parasxos:
Adds `o6_clang_default_design`: builds the default design under **clang** against the **open62541** backend, runs the server, and diffs the address space against the reference. It runs in parallel with the existing gcc jobs and continuously guards the clang / glibc-free hardening (OPCUA-3341) plus the open62541-compat amalgamation fix (OPCUA-3384) instead of relying on one-off local proofs.

Mirrors `o6_default_design` with `CC=clang CXX=clang++` and a `dnf install -y clang` step. AlmaLinux 9 (the `quasar-uasdk:latest` image) ships clang 21.1.8, and the v1.4.6 open62541-compat pin no longer compiles the vendored amalgamation with `-Werror`, so the clang build links end to end. The generated address space is compiler-independent, so it diffs against the same `test_default_quasar_design/reference_ns2.xml` as the gcc job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,27 @@ jobs:
           --open62541_compat_branch ${{ env.OPEN62541_COMPAT_VERSION }}
           --compare_with_nodeset .CI/test_cases/test_default_quasar_design/reference_ns2.xml
 
+  o6_clang_default_design:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/quasar-team/quasar-uasdk:latest
+      credentials:
+        username: ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


parasxos:
Closing: this surfaced that AlmaLinux 9's stock Boost 1.75 does not compile under clang 21 (boost/mpl/aux_/integral_wrapper.hpp non-constant-expression), failing uavariant.cpp and CalculatedVariablesEngine.cpp. gcc 11 is fine. The clang+open62541 path is proven on newer-Boost toolchains; a clang CI job needs a newer-Boost image, not the stock quasar-uasdk:latest. Tracking separately.

PR #381, opened 2026-07-13
correct xsd.exe generator path on windows

ben-farnham:
https://its.cern.ch/jira/browse/OPCUA-3456


parasxos (approved):
Good catch, sorry for the breakage. The WIN32 branch was mine, I assumed a stock CodeSynthesis install and did not know about the xsdcxx rename on your Windows machines. Your version is better anyway, one default everywhere. Merged, tell me if you want a 2.1.1 tag right away.

